PR #29, opened 2021-07-19
Fix callback for nested `signIn` or `addScopes`

### LINKED ISSUE #24: `addScopes` callback not called when nested
Hi, I'm working on updating `GoogleSignIn` to the latest version and facing a problem regarding the `addScopes` function.

Since we are now not able to Sign In with the necessary scopes, I am trying to add them as soon as the user is signed in.
It turns out that by chaining both methods, the callback I send to `addScopes` is never executed.

```
GIDSignIn.sharedInstance.signIn(with: configuration, presenting: presentingViewController) { user, error in
    
    guard error == nil else {
        completion(user, error)
        return
    }
                    
    GIDSignIn.sharedInstance.addScopes(
        scopes,
        presenting: presentingViewController,
        callback: {
            // This callback is never executed
            print(">>> addScopes callback")
            completion($0, $1)
        }
    )
}
```

Testing it manually I'm able to see:

1. The web view being presented for the regular sign in and being dismissed when it succeeds
2. The web view for allowing permissions to be added and dismissed when tapping "Allow"

After these two steps, nothing happens 😔 Am I missing something?

In my opinion the optimal solution would be being able to supply the scopes directly to the `signIn` method so that both tasks could be done in one step instead of two (like suggested on #23) but I wanted to flag this finding I'm running into as well.

Thanks for the great work!

AntonKozlovskyi:
Fixes #24

### DIFF
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -690,10 +690,11 @@ static const NSTimeInterval kMinimumRestoredAccessTokenTimeToExpire = 600.0;
   [authFlow addCallback:^() {
     GIDAuthFlow *handlerAuthFlow = weakAuthFlow;
     if (self->_currentOptions.callback) {
-      dispatch_async(dispatch_get_main_queue(), ^{
-        self->_currentOptions.callback(self->_currentUser, handlerAuthFlow.error);
+        GIDSignInCallback callback = [self->_currentOptions.callback copy];
         self->_currentOptions = nil;
-      });
+        dispatch_async(dispatch_get_main_queue(), ^{
+            callback(self->_currentUser, handlerAuthFlow.error);
+        });
     }
   }];
 }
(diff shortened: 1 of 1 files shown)


AntonKozlovskyi:
@googlebot I signed it!

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -690,10 +690,11 @@ - (void)addCompletionCallback:(GIDAuthFlow *)authFlow {
   [authFlow addCallback:^() {
     GIDAuthFlow *handlerAuthFlow = weakAuthFlow;
     if (self->_currentOptions.callback) {
-      dispatch_async(dispatch_get_main_queue(), ^{
-        self->_currentOptions.callback(self->_currentUser, handlerAuthFlow.error);
+        GIDSignInCallback callback = [self->_currentOptions.callback copy];
```
petea, line 693: Should be able to do without the copy here.

petea (changes requested):
Thanks, Anton!  Could you also apply this fix to the two other locations in GIDSignIn.m where we set "_currentOptions = nil" after executing the callback block?

GoogleSignIn/Sources/GIDSignIn.m
petea, line 693: Use two space tabs.

AntonKozlovskyi:
done

petea (approved):

tgun1370 (commented):
`AntonKozlovskyi:Fix_missing_completion_handler_on_add_scopes`

Issue #23, opened 2021-07-14
SignIn with Scope directly 


floriangbh:
Hello,

Since the 6.0.0 it looks like isn't anymore possible to SignIn with scopes. The only way I'v found is : 
- firstly call `signInWithConfiguration:presentingViewController:callback:`
- then call `addScopes:presentingViewController:callback:`

In my case, I need to connect my user to their Youtube account. It results in two" sign in" in a row, which can be weird for some users.  

Is there a way to make one sign in with scope or this is a volunteer behavior? 

Btw, thanks for the SPM and M1 compatibility! 

Thanks 
Florian

ivopintodasilva:
I'm facing the exact same problem. 

Previously my login flow (login + grant scopes) was contained in a single web view step. Now we need to show one web view for the sign in and another for granting permissions which feels less convenient.

And thanks a lot for SPM and Apple Silicon support! It's awesome 😃

petea:
You are correct, with the 6.0.0 release we're requesting only the basic profile scopes (email, profile, openid) in the user's initial sign-in flow.  If you require an additional scope, this will need to happen via `addScopes:presentingViewController:callback:` sometime after the sign-in flow has been completed.  In keeping with the [incremental authorization](https://developers.google.com/identity/protocols/oauth2/web-server#incrementalAuth) pattern, we recommend that any additional scope requests happen right before the scope is to be used and with context provided to the user as to why they might want to grant access.

floriangbh:
Thanks @petea for the explanation.

petea:
Reopening for discussion.

AntonKozlovskyi:
Hi @petea. Very often we need to get all the scopes in time of sign in(or just after), and in this case it's very inconvenient to user to see `Are you sure you want to sign in` question, make a login and after that see this message again. Some of users could just cancel second one or decide that it's an issue. And in cases when user already gave this permissions( like when he used to use application, removed it and now installed again), after second question it just showing WebView and hide it immediately. It's working, but not very nice looking.

Other approach is just add possibility to hide 2nd popup with sign in question for cases when adding scopes going just after sign in.

mamilov:
Hi @petea. In our App, we have a use case that suffers from this decision. Our users are able to link their calendars (multiple options available) and he only needs sign-in to when we need his calendar consent. This now leads us to having to move the user through two sign-ins (from the eyes of the user) which is bad UX. 
[...]

PR #30, opened 2021-07-19
Add the option to choose additional scopes at sign-in

AntonKozlovskyi:


artemHrebinikChisw:
Very useful functionality. Today I ran into this problem on the project.
How long does it usually take to review such PR?

petea:
Thanks for the PR.  Please note that the separation of `signIn` and `addScopes` is intentional with this release, as mentioned in #23.  Feel free to discuss relevant use cases and impact.

ivopintodasilva:
Any idea on the timeline to conclude if this change will be approved or not?
This is preventing us from updating GoogleSignIn to the latest version...

dpanzer:
@petea It seems that the use cases for this functionality have been documented in #23. I would add my predicament as an SDK developer is that I don't know whether to release an update with a breaking change based on the new 2-step add scope flow, or whether this PR will get approved and I can release an update where the upgrade to v6 will be transparent to users of my SDK (obviously strongly prefer the latter, but overall just want to know what the verdict will be). Thanks!

reslear:
any news?

foxware00:
@petea To weigh in, it's also paramount for us that we don't prompt twice give the first screen of our app uses the calendar permission. I think a lot of people have outlined their concerns and use cases in the issue below. It provides a poor user experience and also means we won't be upgrading to this library in the meantime.

I'm not entirely sure this library should be opinonated on UX flow. If it's required that apps request scope at time of using that feature, surely that is up to the individual use cases. Those apps which require scope extensions immediately are penalised.

petea:
@AntonKozlovskyi thanks again for your contribution!  We now plan to ship this with the 6.2 release.  See my [update](https://github.com/google/GoogleSignIn-iOS/issues/23#issuecomment-1061230684) on #23.

petea (approved):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -192,6 +192,20 @@ - (void)signInWithConfiguration:(GIDConfiguration *)configuration
   [self signInWithOptions:options];
 }
 
+- (void)signInWithConfiguration:(GIDConfiguration *)configuration
```
etrdotcom, line 220: 000IOEC. P kikx.ios@yahoo.c omreceipt.here

PR #31, opened 2021-07-20
Addition of promotional prompt strings to `en.lproj/GoogleSignIn.strings`

thomasmburke:
There are some strings missing in `en.lproj/GoogleSignIn.strings` inside GoogleSignIn.bundle which is part of the iOS Google sign-in SDK, so adding them in to avoid crashes in apps that use them.

These strings already exist in `en_GB.lproj/GoogleSignIn.strings`, so just making them available in `en.lproj/GoogleSignIn.strings` as well.


petea:
Luckily these strings are no longer in use and can be removed from the other locales, as they were from en.lproj.

petea:
Thanks for the heads up here.  Closing in lieu of #36.

PR #34, opened 2021-07-21
Prep 6.0.1 release

petea:


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 6.0.1 (2021-7-20)
```
paulb777: FYI, we either avoid dates in release notes or don't go more granular than a month - since they're easy to get wrong, can change before release, and git history is the real source of truth.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
...
+- Fixes nested callbacks not being called for signIn and addScopes methods.
```
paulb777, line 2: Consider including issue numbers for the issues fixed in this patch.

paulb777 (approved):

CHANGELOG.md
petea: I'll see how much trouble it ends up being.  I do like having a detailed release record that can be used outside of GitHub.

ch40w31 (approved):

Issue #33, opened 2021-07-20
Add "pod lib lint --use-libraries" to GitHub actions to validate GoogleSignIn builds as a library


jmagman:
Currently there's a GitHub Action to test GoogleSignIn builds as a framework.  Add a test to validate it still builds as a library.
https://github.com/google/GoogleSignIn-iOS/blob/5193a5befc1304c67201cc80f7d2818a0f8aadf4/.github/workflows/tests.yml#L22-L23
```yaml
    - name: Lint podspec using local source as library
      run: pod lib lint --verbose --allow-warnings --use-libraries
```

paulb777:
Here's an example that also tests `--use-static-framework` as well as other Apple platforms - https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/core.yml#L68

jmagman:
https://github.com/firebase/firebase-ios-sdk/blob/04ab074265633049d5518866515e3bd7fc0de246/.github/workflows/core.yml#L78-L79
says
```yml
          # Tests are skipped since the Swift tests need modules.
          '--skip-tests --use-libraries'
```
"Tests are skipped since the Swift tests need modules" is the same bug in that repo as https://github.com/google/GoogleSignIn-iOS/issues/32.  I filed this GitHub Action request to validate the modules are generated for libraries when that issue is fixed, which can be done by adding an `@import GoogleSignIn ` in the unit tests or making any of them Swift files.

paulb777:
That's arguably not a bug since it can be argued that the decision to use modules or not should be left to the pod consumer who can specify `use_frameworks! ` or `use_frameworks! :linkage => :static` to generate frameworks and modules if desired. I believe CocoaPods has always documented that Swift clients should use `use_frameworks! ` and it was changed awhile ago so that the `Podfile` created by `pod init` has `use_frameworks!` on.  I realize that I'm treating modules and frameworks as a synonym, and maybe that's an issue, but we haven't heard about it from Firebase clients.

jmagman:
Indeed, Swift clients should use `use_frameworks!`.  I investigated this more in https://github.com/flutter/flutter/issues/25676#issuecomment-532464440.  However there's no reason not to support module maps regardless of the client language.  `DEFINES_MODULE=YES` is the default build setting in a new vanilla Xcode Objective-C or Swift frameworks target.

In any case for the sake of this GitHub Action request, `--use-libraries` is still a good thing to test for Objective-C clients.

[...]

PR #53, opened 2021-08-19
Always generate a clang module.

### LINKED ISSUE #32: GoogleSignIn does not define a clang module, cannot @import GoogleSignIn when built as a library
`GoogleSignIn` does not generate a clang module when built as a library instead of a framework.  This forces users to use `use_frameworks!` in their Podfile, which isn't the default.

### Steps to Reproduce
1. Remove `use_frameworks!` from Sample Podfile to generate as a library
https://github.com/google/GoogleSignIn-iOS/blob/954652d2e3f87218e3fd81bc0cffe88b0b226cce/Sample/Podfile#L2
2. Open SignInSampleForPod.xcworkspace and build.

Compilation failure: 
```
GoogleSignIn-iOS/Sample/Source/SignInViewController.h:19:9: Module 'GoogleSignIn' not found
```

Screen Shot 2021-07-20 at 11 06 43 AM

### To fix
Add  `'DEFINES_MODULE' => 'YES'` to `pod_target_xcconfig`
https://github.com/google/GoogleSignIn-iOS/blob/5193a5befc1304c67201cc80f7d2818a0f8aadf4/GoogleSignIn.podspec#L38-L41
becomes:
```ruby
 s.pod_target_xcconfig = { 
   'GCC_PREPROCESSOR_DEFINITIONS' => 'GID_SDK_VERSION=' + s.version.to_s, 
   'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"',
   'DEFINES_MODULE' => 'YES'
 } 
```
See "Modular Headers" section of https://blog.cocoapods.org/CocoaPods-1.5.0/

petea:
Allow CocoaPods developers to use modules without requiring that `use_frameworks!` or `:modular_headers => true` be used their Podfile.

Resolves #32 and #33.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,10 +16,10 @@ jobs:
     strategy:
       matrix:
         flags: [
-	  '',
-	  '--use-libraries',
-	  '--use-static-frameworks'
-	]
+          '',
+          '--use-libraries',
+          '--use-static-frameworks'
+        ]
     steps:
     - uses: actions/checkout@v2
     - name: Update Bundler
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,6 +13,13 @@ jobs:
 
   pod-lib-lint:
     runs-on: macOS-latest
+    strategy:
+      matrix:
+        flags: [
+	  '',
+	  '--use-libraries',
+	  '--use-static-frameworks'
+	]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


paulb777 (approved):
I recommend adding two more `pod lib lint` commands to the CI yml file to make sure other library/module variations continue to build.  With the option `--use-libraries` and the option `--use-static-frameworks`

ch40w31 (approved):

PR #64, opened 2021-09-29
Add default calendar

cmlarsen:


petea:
Thanks for the PR.  For our purposes here, it's important that the default scopes be the bare minimum needed to authenticate the user.

PR #67, opened 2021-11-05
allow additional scopes at signin

feighter09:
Right now you have to log in and you're granted Profile & Email scopes, but then you have to go through the flow again to request additional scopes for email, calendar. Because we sign in with Google in order to link to your calendar, this results in a super annoying double-login prompt..
Adding functionality to request these additional scopes when you first sign in


enrico-querci:
🆙

petea:
Thanks for your PR.  We have merged #30 which should address the same problem.

PR #70, opened 2021-11-09
Set include_granted_scopes=true in the auth request

petea:
In order to restore a user's previously granted scopes at sign-in time, we'll want to start using include_granted_scopes=true in the auth request.


mdmathias (changes requested):
Could we add a test to `GIDSignInTest.m` to ensure that the `additionalParameters` passed to `-[OIDAuthorizationRequest initWithConfiguration:clientId:scopes:redirectURL:responseType:additionalParameters:]` is correct? It'd be nice to document in test the expectations we have for this.

petea:
> Could we add a test to `GIDSignInTest.m` to ensure that the `additionalParameters` passed to `-[OIDAuthorizationRequest initWithConfiguration:clientId:scopes:redirectURL:responseType:additionalParameters:]` is correct? It'd be nice to document in test the expectations we have for this.

Added a check for the static include_granted_scopes value.  The other additionalParameters values are dynamic and get checked in specific tests.

mdmathias (approved):

PR #78, opened 2021-12-14
Mark signIn and addScopes methods as unavailable in app extensions.

petea:
To avoid a potential sharedApplication reference in AppAuth for an extension target, which would generate a runtime error, we will mark these methods as unavailable to extensions and catch any references at compile time.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
@@ -104,7 +104,8 @@ typedef void (^GIDDisconnectCallback)(NSError *_Nullable error);
 ///     called asynchronously on the main queue.
 - (void)signInWithConfiguration:(GIDConfiguration *)configuration
        presentingViewController:(UIViewController *)presentingViewController
-                       callback:(nullable GIDSignInCallback)callback;
+                       callback:(nullable GIDSignInCallback)callback
+    NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.");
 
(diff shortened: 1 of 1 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -139,7 +141,8 @@ typedef void (^GIDDisconnectCallback)(NSError *_Nullable error);
 ///     called asynchronously on the main queue.
 - (void)addScopes:(NSArray<NSString *> *)scopes
     presentingViewController:(UIViewController *)presentingViewController
-                    callback:(nullable GIDSignInCallback)callback;
+                    callback:(nullable GIDSignInCallback)callback
+    NS_EXTENSION_UNAVAILABLE("The add scopes flow is not supported in App Extensions."); 
 
 /// Marks current user as being in the signed out state.
 - (void)signOut;
```
mdmathias, line 148: Does it make sense that `-signOut` and `-disconnectWithCallback:` are available?

mdmathias (approved):
Just one question about marking other methods as unavailable; LGTM otherwise.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
petea: Technically, those methods are available.  Whether or not they make sense to use in an extension, I'm thinking we leave to developers.

mdmathias: Sounds like the issue at stake here doesn't impact these methods (since they do not use the `sharedApplication` to do their work). I'm resolving.

PR #87, opened 2022-01-04
Replace deprecated API calls

XiangtianDai:
Replaces deprecated APIs with new ones on later iOS versions to avoid compiler warnings.

### DIFF
--- a/GoogleSignIn/Sources/GIDEMMErrorHandler.m
+++ b/GoogleSignIn/Sources/GIDEMMErrorHandler.m
@@ -99,9 +99,17 @@ typedef enum {
       completion();
       return;
     }
-    CGRect keyWindowBounds = CGRectIsEmpty(keyWindow.bounds) ?
+    UIWindow *alertWindow;
+    if (@available(iOS 13, *)) {
+      if (keyWindow.windowScene) {
+        alertWindow = [[UIWindow alloc] initWithWindowScene:keyWindow.windowScene];
+      }
+    }
+    if (!alertWindow) {
+      CGRect keyWindowBounds = CGRectIsEmpty(keyWindow.bounds) ?
         keyWindow.bounds : [UIScreen mainScreen].bounds;
-    UIWindow *alertWindow = [[UIWindow alloc] initWithFrame:keyWindowBounds];
+      alertWindow = [[UIWindow alloc] initWithFrame:keyWindowBounds];
+    }
     alertWindow.backgroundColor = [UIColor clearColor];
     alertWindow.rootViewController = [[UIViewController alloc] init];
     alertWindow.rootViewController.view.backgroundColor = [UIColor clearColor];
--- a/GoogleSignIn/Sources/GIDEMMErrorHandler.m
+++ b/GoogleSignIn/Sources/GIDEMMErrorHandler.m
@@ -15,11 +15,11 @@
 
 #if TARGET_OS_IOS
 
-#import "GoogleSignIn/Sources/GIDEMMErrorHandler.h"
+#import "third_party/objective_c/GoogleSignIn/Sources/GIDEMMErrorHandler.h"
 
 #import <UIKit/UIKit.h>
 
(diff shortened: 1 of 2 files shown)


GoogleSignIn/Tests/Unit/GIDEMMErrorHandlerTest.m
```diff
@@ -66,6 +66,11 @@ - (void)setUp {
   _isIOS10 = [UIDevice currentDevice].systemVersion.integerValue == 10;
   _keyWindowSet = NO;
   _presentedViewController = nil;
+  UIWindow *fakeKeyWindow = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+  [GULSwizzler swizzleClass:[GIDEMMErrorHandler class]
```
petea, line 70: Should we test -keyWindow internals rather than mocking?

XiangtianDai: I couldn't find a way to effectively test it (i.e., not just repeat what the code does), because the correctness of this method really depends on the return value of the system APIs -- any suggestions?

petea: You could mock connectedScenes/windows/keyWindow and check that window selection happens as expected.  May not be worth it, I leave it to you.

GoogleSignIn/Sources/GIDEMMErrorHandler.m
```diff
@@ -90,7 +90,12 @@ - (BOOL)handleErrorFromResponse:(NSDictionary<NSString *, id> *)response
...
   // All UI must happen in the main thread.
   dispatch_async(dispatch_get_main_queue(), ^() {
-    UIWindow *keyWindow = [UIApplication sharedApplication].keyWindow;
+    UIWindow *keyWindow = [self keyWindow];
+    if (!keyWindow) {
+      // Shouldn't happen, just in case.
+      completion();
+      return;
+    }
     CGRect keyWindowBounds = CGRectIsEmpty(keyWindow.bounds) ?
```
lzw-all-in, line 99: Hi @XiangtianDai , Could we also support scene-based lifecycle here ?  Which need connect UIWindow to a specific UIWindowScene.

Here is an example of my PR #92

GoogleSignIn/Tests/Unit/GIDEMMErrorHandlerTest.m
XiangtianDai: Added a test for the `keyWindow` method by mocking its behavior. Not feeling very good about it, but hopefully better than nothing.

GoogleSignIn/Sources/GIDEMMErrorHandler.m
```diff
@@ -130,6 +135,33 @@ - (BOOL)handleErrorFromResponse:(NSDictionary<NSString *, id> *)response
   return YES;
 }
 
+// This method is exposed to the unit test.
+- (UIWindow *)keyWindow {
```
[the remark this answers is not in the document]
XiangtianDai: Done.

GoogleSignIn/Sources/GIDEMMErrorHandler.m
XiangtianDai: @lzw-all-in Updated the code to use `initWithWindowScene:` if available -- does this look good to you?

petea:
@XiangtianDai are you ready to merge this?

GoogleSignIn/Sources/GIDEMMErrorHandler.m
petea: @XiangtianDai are we ready to resolve this thread?

XiangtianDai: I believe my updated code addresses comments from @lzw-all-in

petea (approved):

PR #92, opened 2022-01-21
Fix alert window disappearance issue under scene-based lifecycle in GIDEMMErrorHandler

lzw-all-in:
## PR Statement
Hi reviewers:

My app is currently using `GoogleSignIn` to support google account. However, I intend to migrate my app from app-based lifecycle to scene-based lifecycle(Which is announced by Apple at WWDC19). When I go through all the third-party dependencies, I found `GoogleSignIn/GIDEMMErrorHandler.m` have a potential risk under scene-based lifecycle: The error alert will never show up because alertWindow was not connected to UIWindowScene once the app adopt scene-based lifecycle. In this PR, I'm trying to make a minimum changes to support scene-based lifecycle.

## Future Improvement
For simplicity, my changes only support single-scene app and for multi-scene app it might show up the alertWindow at a wrong scene. But It should work fine in most cases. According to my understanding as a third-party contributor, support multi-scene in GoogleSignIn will be a heavy work. Because you need you store the UIWindowScene user interact with to login its account and pass it to initialize alertWindow at a proper time.

Looking forward to see your response.

Thanks.

### DIFF
--- a/GoogleSignIn/Sources/GIDEMMErrorHandler.m
+++ b/GoogleSignIn/Sources/GIDEMMErrorHandler.m
@@ -95,22 +95,22 @@ typedef enum {
     // Compatible with different iOS lifecycles for UIWindow initialization.
     if (@available(iOS 13.0, *)) {
       for (UIWindow *window in [UIApplication sharedApplication].windows) {
-          if (window.isKeyWindow) {
-              keyWindow = window;
-              break;
-          }
+        if (window.isKeyWindow) {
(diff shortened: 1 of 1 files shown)


lzw-all-in:
Hi @petea Could you help me to review it ? Thanks s o much.

petea:
@lzw-all-in is this now addressed in #87?

petea:
I believe this has now been addressed as part of #87.

PR #97, opened 2022-02-01
Add macOS support version in Package.swift

Alex-4-Git:
We can support macOS 10.15 and above for now.


Al-Kuka:
Hi,
please help me:
I just added package to own project and got this error:
The package product 'GTMAppAuth' requires minimum platform version 10.11 for the macOS platform, but this target supports 10.10
How to resolve it?

Thanks!

Alex-4-Git:
> 
That's because GSI doesn't reconcile these. After I merge this changes these errors will go away. Right now GSi doesn't support macOS, but it is coming soon.

ch40w31 (approved):

PR #102, opened 2022-02-26
GSI macOS support preparation

Alex-4-Git:
1. Edit Cocoapods and Swift Package manager config file.
2. Annotate iOS specific components.

### DIFF
--- a/GoogleSignIn.podspec
+++ b/GoogleSignIn.podspec
@@ -26,9 +26,9 @@ The Google Sign-In SDK allows users to sign in with their Google account from th
     'CoreText',
     'Foundation',
     'LocalAuthentication',
-    'Security',
-    'UIKit'
+    'Security'
   ]
+  s.ios.framework = 'UIKit'
   s.dependency 'AppAuth', '~> 1.4'
   s.dependency 'GTMAppAuth', '~> 1.0'
   s.dependency 'GTMSessionFetcher/Core', '~> 1.1'
--- a/Package.swift
+++ b/Package.swift
@@ -22,7 +22,10 @@ let googleSignInVersion = "6.1.0"
 let package = Package(
   name: "GoogleSignIn",
   defaultLocalization: "en",
-  platforms: [.iOS(.v9)],
+  platforms: [
+    .macOS(.v10_15),
+    .iOS(.v9)
+  ],
   products: [
     .library(
       name: "GoogleSignIn",
@@ -75,7 +78,7 @@ let package = Package(
         .linkedFramework("Foundation"),
         .linkedFramework("LocalAuthentication"),
         .linkedFramework("Security"),
-        .linkedFramework("UIKit"),
+        .linkedFramework("UIKit", .when(platforms: [.iOS])),
       ]
     ),
     .testTarget(
--- a/GoogleSignIn/Sources/GIDSignInCallbackSchemes.m
+++ b/GoogleSignIn/Sources/GIDSignInCallbackSchemes.m
@@ -14,8 +14,6 @@
 
 #import "GoogleSignIn/Sources/GIDSignInCallbackSchemes.h"
 
-#import <UIKit/UIKit.h>
-
 NS_ASSUME_NONNULL_BEGIN
 
 @implementation GIDSignInCallbackSchemes {


GoogleSignIn.podspec
```diff
@@ -26,9 +26,10 @@ The Google Sign-In SDK allows users to sign in with their Google account from th
     'CoreText',
     'Foundation',
     'LocalAuthentication',
-    'Security',
-    'UIKit'
+    'Security'
   ]
+  s.ios.framework = 'UIKit'
+  s.osx.framework = 'AppKit'
```
petea, line 32: Are we using AppKit on macOS?

Package.swift
```diff
@@ -75,7 +78,8 @@ let package = Package(
         .linkedFramework("Foundation"),
         .linkedFramework("LocalAuthentication"),
         .linkedFramework("Security"),
-        .linkedFramework("UIKit"),
+        .linkedFramework("AppKit", .when(platforms: [.macOS])),
```
petea, line 81: Same as above, are we using AppKit on macOS?

GoogleSignIn/Sources/GIDEMMErrorHandler.h
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <TargetConditionals.h>
```
mdmathias, line 16: Is there a reason to use `#include` over `#import` here and throughout?

mdmathias (approved):
LGTM. I do suggest using `#import` if there is no reason to use `#include`.

GoogleSignIn/Sources/GIDEMMErrorHandler.h
```diff
@@ -13,6 +13,9 @@
...
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 18: Currently, EMM support is disabled for Mac Catalyst, so we should be able to exclude that case here.  See #77 for details.

We should update the condition used in #77 to only enable EMM support for iOS.

GoogleSignIn/Sources/GIDEMMErrorHandler.m
```diff
@@ -11,6 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 16: Same as above.

GoogleSignIn/Sources/GIDMDMPasscodeCache.h
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 18: Same as above.

GoogleSignIn/Sources/GIDMDMPasscodeCache.m
```diff
@@ -11,6 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 16: Same as above.

GoogleSignIn/Sources/GIDMDMPasscodeState.h
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 18: Same as above.

GoogleSignIn/Sources/GIDMDMPasscodeState.m
```diff
@@ -11,6 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 16: Same as above.

GoogleSignIn/Sources/GIDMDMPasscodeState_Private.h
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <TargetConditionals.h>
+
+#if TARGET_OS_IOS || TARGET_OS_MACCATALYST
```
petea, line 18: Same as above.

petea (changes requested):

GoogleSignIn/Sources/GIDEMMErrorHandler.h
ch40w31, line 18: Instead of conditionally including everything in the header, can we conditionally import this and other EMM headers where they are used?

Bonus point if we can conditionally include the EMM files at the dependency manager level (podspec and Package.swift) too.

GoogleSignIn.podspec
Alex-4-Git: Yes, we need to use NSWindow.

Package.swift
Alex-4-Git: Yes, we need to use NSWindow.

GoogleSignIn/Sources/GIDEMMErrorHandler.h
Alex-4-Git: These headers are only imported when they are needed in the next PR. I see your point here. I will revert this PR.

Alex-4-Git: I already did that in the next PR. I guess I can revert this PR.

GoogleSignIn/Sources/GIDEMMErrorHandler.h
Alex-4-Git: I am following conventions in Google 3 and other libraries. Do you think #import is better here?

ch40w31: This specific header is a C header so `#include` is appropriate.

Package.swift
Alex-4-Git: Move the changes to the next PR.

GoogleSignIn/Sources/GIDEMMErrorHandler.m
Alex-4-Git: Removed.

GoogleSignIn/Sources/GIDMDMPasscodeCache.h
Alex-4-Git: Removed.

GoogleSignIn/Sources/GIDEMMErrorHandler.h
mdmathias: In reading up on this, `#import` should be fine and perhaps even helpful as it obviates the need for include guards. In general, using `#include` may entail a small performance hit in compilation as it requires that the header is opened and the include guards (if present) are read and respected. Thus, using `#include` ensures that the importer respects the author's intention at the potential cost of performance. 

Nonetheless, it doesn't seem like an issue in this case as the header does have an include guard. :)

GoogleSignIn/Sources/GIDMDMPasscodeCache.m
Alex-4-Git: Removed.

GoogleSignIn/Sources/GIDMDMPasscodeState.h
Alex-4-Git: Removed.

GoogleSignIn/Sources/GIDMDMPasscodeState.m
Alex-4-Git: Removed.

GoogleSignIn/Sources/GIDMDMPasscodeState_Private.h
Alex-4-Git: Removed.

GoogleSignIn.podspec
Alex-4-Git: Moved this to the next PR.

GoogleSignIn/Sources/GIDEMMErrorHandler.h
Alex-4-Git: Ack

Alex-4-Git: Reopen this PR to only modify configuration.

petea (approved):

Issue #49, opened 2021-08-16
Using v6 in SwiftUI: `.signIn()` workaround and suggestions for SwiftUI support


peterkos:
## Background

With SwiftUI being the way forward for iOS dev, I would have hoped GoogleSignIn had better support for SwiftUI. 

Unfortunatley there isn't any SwiftUI-focused documentation, and the patterns changed in v6 (e.g., no delegates) seem to make implementing GIDSignIn harder in SwifUI.

The biggest issue I've had (and clearest example) is below:

## Sign In

The main issue I've run into is with this function:
```swift
- (void)signInWithConfiguration:(GIDConfiguration *)configuration
       presentingViewController:(UIViewController *)presentingViewController
                       callback:(nullable GIDSignInCallback)callback;
```

SwiftUI doesn't have view controllers, but we need to instantiate _a_ sign-in view from SwiftUI.

There's two workarounds:
1. Create a `UIViewRepresentable` view that accesses the root view controller with something like `UIApplication.shared.windows.last?.rootViewController`
1. Create a `UIViewRepresentable` view that instantiates a dummy `UIViewController` that then calls some `AuthManager` singleton, which in turn calls `sharedInstance.signIn(...)`.

This is so confusing to implement primarily because there _already existed_ a delegate pattern and a non-`presentingVC` parameter [before v6 of this library](https://developers.google.com/identity/sign-in/ios/quick-migration-guide).

Tying it all together would be something like this:

```swift
// AuthManager.swift

class AuthManager {

    var shared = AuthManager()
    var config = GIDConfiguration()

    func signIn(withVC vc: UIViewController) {
        GIDSignIn.signIn(with: config, presenting: vc) { /* ... */ }
    }
}


// AuthWrapper.swift 

class AuthVC: UIViewController {
    func signIn() {
        AuthManager.shared.signIn(withVC: self)
    }
}

class AuthVCWrapper: UIViewRepresentable {
    let authVC = AuthVC() // strong ref
    func makeUIViewController() {
        return authVC
    }
    func signIn() {
        authVC.signIn() // single responsibility principle
    }
}


// SignIn.swift

class SignIn: View {
    var authWrapper = AuthVCWrapper()

    var body: some View {
        ZStack {
            authWrapper
            Button(action: authWrapper.signIn()) {
                Text("Sign In")
            }
        }
    }
}
```

To me personally this seems verbose. 

## Delegates

[...]

PR #103, opened 2022-02-26
Add a native SwiftUI button

mdmathias:
This change adds a native SwiftUI button, and seeks to provide further native Swift support (#49).

This PR will also lay the foundation for more Swift-specific development by establishing a separate library.


Package.swift
```diff
@@ -17,7 +17,7 @@
 
 import PackageDescription
 
-let googleSignInVersion = "6.1.0"
+let googleSignInVersion = "6.2"
```
petea, line 20: Can we wait on the version bump?  I usually do this all at once in a separate PR.

mdmathias: Sure thing.

GoogleSignIn/Sources/GIDSignInButtonStyling.m
```diff
@@ -0,0 +1,116 @@
...
+const NSUInteger kNumGIDSignInButtonStyleColors = 2;
+
+@interface GIDSignInButtonStyling ()
+
+@end
+
+@implementation GIDSignInButtonStyling
+
++ (NSURL *)googleIconURL {
+  NSURL *imgURL = [SWIFTPM_MODULE_BUNDLE URLForResource:kGoogleImageName
```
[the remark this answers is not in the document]
mdmathias: Good point. Fixed.

petea:
Look into using the --include-podspecs flag for pod lib lint for loading the GoogleSignIn podspec from the local repo.

.github/workflows/tests.yml
```diff
@@ -15,19 +15,25 @@ jobs:
...
-          '--use-static-frameworks'
+        podspec: [GoogleSignIn.podspec, GoogleSignInSwift.podspec]
+        flag: [
+          "", 
+          "--use-libraries", 
+          "--use-static-frameworks"
         ]
+        include:
+          - podspec: GoogleSignInSwift.podspec
+            includePodspecFlag: "--include-podspecs='*.podspec'"
```
petea, line 26: Now that we're only targeting the Swift pod run, we could be more specific:

`--include-podspecs='GoogleSignIn.podspec'`

GoogleSignInSwift.podspec
```diff
@@ -0,0 +1,29 @@
...
+  s.osx.deployment_target = macos_deployment_target
+  s.prefix_header_file = false
+  s.source_files = [
+    'GoogleSignInSwift/Sources/*.swift',
+  ]
+  s.frameworks = [
+    'CoreGraphics',
+    'SwiftUI',
+  ]
+  s.dependency 'GoogleSignIn', '~> 6.1'
```
petea, line 28: Update this to 6.2 after rebasing from main to bring in my 6.2 version bump.

GoogleSignInSwift.podspec
```diff
@@ -0,0 +1,29 @@
+Pod::Spec.new do |s|
+  s.name = 'GoogleSignInSwift'
+  s.version = '0.1.0'
```
petea, line 3: I think we should sync versions with the main pod, setting this to 6.2.0.  Although we could have independent versioning for the pods, we couldn't do this for the SPM products because SPM uses one git tag to version all products in the package definition.

GoogleSignInSwift.podspec
```diff
@@ -0,0 +1,29 @@
...
+  s.swift_version = '4.0'
+  s.summary = 'Adds Swift-focused support for Google Sign In.'
+  s.description = <<-DESC
+The Google Sign-In SDK allows users to sign in with their Google account from third-party apps.
```
petea, line 7: Let's just sync this text with the summary for now.

GoogleSignIn/Sources/GIDSignInButtonStyling.m
```diff
@@ -0,0 +1,120 @@
+// Copyright 2021 Google LLC
```
petea, line 1: "2022" here and in the other new class files.

GoogleSignIn/Sources/GIDSignInButtonStyling.m
```diff
@@ -0,0 +1,120 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButtonStyling.h"
+#import "GoogleSignIn/Sources/NSBundle+GID3PAdditions.h"
+
```
petea, line 17: NS_ASSUME_NONNULL_BEGIN ?

GoogleSignIn/Sources/GIDSignInButtonStyling.m
```diff
@@ -0,0 +1,120 @@
...
+  return [NSURL URLWithString:iconPath];
+}
+
+/// Retrieves the Google font URL from the bundle.
++ (NSURL *)googleFontURL {
+  NSBundle *bundle = [NSBundle gid_frameworkBundle];
+  NSString *fontPath = [bundle pathForResource:kFontNameRobotoBold ofType:@"ttf"];
+  return [NSURL URLWithString:fontPath];
+}
+
```
petea, line 119: NS_ASSUME_NONNULL_END ?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
```diff
@@ -19,26 +19,11 @@
...
+// Forward declarations of enums
+typedef NS_ENUM(NSInteger, GIDSignInButtonStyle);
+typedef NS_ENUM(NSInteger, GIDSignInButtonColorScheme);
 
-/// The color schemes supported by the `GIDSignInButton`.
-typedef NS_ENUM(NSInteger, GIDSignInButtonColorScheme) {
-  kGIDSignInButtonColorSchemeDark = 0,
-  kGIDSignInButtonColorSchemeLight = 1
-};
+NS_ASSUME_NONNULL_BEGIN
```
petea, line 26: Move this back above typedefs for consistency?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
```diff
@@ -19,26 +19,11 @@
...
-/// The following dimensions (in points) fit for all languages:
-/// - kGIDSignInButtonStyleStandard: 230 x 48
-/// - kGIDSignInButtonStyleWide:     312 x 48
-/// - kGIDSignInButtonStyleIconOnly: 48 x 48 (no text, fixed size)
-typedef NS_ENUM(NSInteger, GIDSignInButtonStyle) {
-  kGIDSignInButtonStyleStandard = 0,
-  kGIDSignInButtonStyleWide = 1,
-  kGIDSignInButtonStyleIconOnly = 2
-};
+// Forward declarations of enums
```
petea, line 22: If we don't define the enum values here, is this not a breaking change?

petea (changes requested):

GoogleSignInSwift.podspec
mdmathias: Sounds good.

GoogleSignInSwift.podspec
mdmathias: Okay, I'll set this to 6.2.0. One great leap forward!

GoogleSignInSwift.podspec
mdmathias: Done

GoogleSignInSwift.podspec
```diff
@@ -0,0 +1,27 @@
+Pod::Spec.new do |s|
+  s.name = 'GoogleSignInSwift'
+  s.version = '6.2.0'
+  s.swift_version = '4.0'
+  s.summary = 'Adds Swift-focused support for Google Sign In.'
+  s.description = 'Adds Swift-focused support for Google Sign In.'
```
petea, line 6: Looks like I broke your pod lib lint here.  How about we try:

"Additional Swift support for the Google Sign-In SDK."

GoogleSignInSwift.podspec
```diff
@@ -0,0 +1,27 @@
+Pod::Spec.new do |s|
+  s.name = 'GoogleSignInSwift'
+  s.version = '6.2.0'
+  s.swift_version = '4.0'
+  s.summary = 'Adds Swift-focused support for Google Sign In.'
```
petea, line 5: Don't forget the hyphen: "Google Sign-In"

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
mdmathias: I've imported `GIDSignInButtonStyling.h` into `GIDSignInButton.h`  to circumvent this. Per offline chat, I agree that this isn't desirable as we're leaking implementation details into our API, but I cannot see a way of creating a Swift target that depends upon this information without doing something like that. We can treat this as a temporary solution and create an issue to track fixing this.

guilhermedfs (approved):

GoogleSignInSwift/Sources/GIDSignInButton_SwiftUI.swift
```diff
@@ -0,0 +1,216 @@
...
+    switch self {
+    case .iconOnly: return Width(min: CGFloat(kIconWidth), max: CGFloat(kIconWidth))
+    case .standard: return Width(min: 90, max: .infinity)
+    case .wide: return Width(min: 170, max: .infinity)
+    default:
+      fatalError("Unrecognized case for `GIDSignInButtonStyle: \(self)")
+    }
+  }
+
+  fileprivate var buttonText: String {
```
petea, line 44: We'll need to load strings from the shared bundle in order to handle localization.

GoogleSignInSwift/Sources/GIDSignInButton_SwiftUI.swift
```diff
@@ -0,0 +1,216 @@
...
+struct SwiftUIButtonStyle: ButtonStyle {
+  let style: GIDSignInButtonStyle
+
+  func makeBody(configuration: Configuration) -> some View {
+    configuration.label
+      .frame(minWidth: style.width.min,
+             maxWidth: style.width.max,
+             minHeight: CGFloat(kButtonHeight),
+             maxHeight: CGFloat(kButtonHeight))
+      .foregroundColor(.blue)
```
petea, line 66: The default text color for the SwiftUI button seems to be blue rather than grey as in GIDSignInButton, is this the cause?

petea (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
petea: Discussed offline and agreed to pursue a refactor with the goal of not having any public API surface impact.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthday.xcodeproj/project.pbxproj
```diff
@@ -384,22 +389,27 @@
...
 /* Begin XCRemoteSwiftPackageReference section */
-		7345AD252703DA3E0020AFB1 /* XCRemoteSwiftPackageReference "GoogleSignIn-iOS" */ = {
+		73F87DE427C9B98A00E12BFE /* XCRemoteSwiftPackageReference "GoogleSignIn-iOS" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/google/GoogleSignIn-iOS.git";
+			repositoryURL = "https://github.com/google/GoogleSignIn-iOS";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 6.0.2;
+				branch = "mdmathias/swiftui-button";
```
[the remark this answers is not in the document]
mdmathias: I'll remove it before submitting.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButtonStyling.h
```diff
@@ -0,0 +1,133 @@
...
+
+// Button text font size.
+extern const CGFloat kFontSize;
+
+#pragma mark - Icon Constants
+
+// The name of the image for the Google "G"
+extern NSString *const kGoogleImageName;
+
+// Keys used for NSCoding.
```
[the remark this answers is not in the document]
mdmathias: Removed the file.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthday.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
petea: Pin's recently merged #111 adds GoogleSignIn as a local package, so you should be able to simply remove this after merging.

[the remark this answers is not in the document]
petea: Pin's recently merged #111 adds GoogleSignIn as a local package, so you should be able to simply remove this after merging.

GoogleSignInSwift/Sources/GIDSwiftUISignInButton.swift
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+import SwiftUI
+import CoreGraphics
+
+// MARK: - Sign-In Button
+
+/// A Google Sign In button to be used in SwiftUI.
+@available(iOS 13.0, macOS 10.15, *)
+public struct GIDSwiftUISignInButton: View {
```
petea, line 24: Do we want the `GID` prefix in the Swift environment?

petea, line 24: It looks like ObjC interop isn't possible for structs.  Are we OK with that?  Would it be possible to use a class here?

petea, line 24: It looks like ObjC interop isn't possible for structs.  Are we OK with that?  Would it be possible to use a class here?

GoogleSignInSwift/Sources/SignInButtonStyling.swift
```diff
@@ -0,0 +1,260 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+#if os(iOS) || targetEnvironment(macCatalyst)
+import UIKit
```
petea, line 20: Where is this being used?

petea, line 20: Where is this being used?

GoogleSignInSwift/Sources/GIDSwiftUISignInButton.swift
mdmathias: Do you have a use-case for this? We have an existing button working in ObjC/UIKit right now, and so using this SwiftUI button over there shouldn't be a need. Moreover, I believe SwiftUI enforces conformers of `View` to be value types. Using a `final class` (e.g., `final class MyView: View {...}`) will yield an error: "Fatal error: views must be value types (either a struct or an enum)."

If we want to expose this button to UIKit, then we'd need to use [`UIHostingController`](https://developer.apple.com/documentation/swiftui/uihostingcontroller) and [`NSHostingView`](https://developer.apple.com/documentation/swiftui/nshostingview).

mdmathias: Resolving as dupe.

GoogleSignInSwift/Sources/SignInButtonStyling.swift
mdmathias: Resolving as a dupe.

mdmathias: Actually, this was silly. I don't believe these imports are needed here!

GoogleSignInSwift/Sources/GIDSwiftUISignInButton.swift
mdmathias: What do you think about renaming to `GoogleSignInButton`

petea: Yes.  I think either `GoogleSignInButton` or `SignInButton`.

With namespacing:

```
GoogleSignInSwift.GoogleSignInButton
or
GoogleSignInSwift.SignInButton
```

and eventually:

```
GoogleSignIn.GoogleSignInButton
or
GoogleSignIn.SignInButton
```

petea: I'm thinking of the AppKit use case.  A traditional ObjC macOS app that wants to use the SwiftUI button.  It's an edge case, but it'd be nice to have a story.  One solution for the developer would be to write a AppKit view in Swift that uses our SwiftUI button via `NSHostingView`.

mdmathias: I went with `GoogleSignInButton`.

mdmathias: Tracking this with [a new issue](https://github.com/google/GoogleSignIn-iOS/issues/117).

petea (approved):

PR #107, opened 2022-03-10
Import GTMFetcherAuthorizationProtocol in GIDAuthentication

### LINKED ISSUE #106: Cannot access fetcherAuthorizer property on GIDAuthentication from swift
The `fetcherAuthorizer` method signature is not visible in swift generated interface of `GIDAuthentication` . 
This prevent the use of this library to work with https://github.com/google/google-api-objectivec-client-for-rest from swift
Screenshot 2022-03-10 at 15 08 22

mdmathias:
Forward declaring the protocol in GIDAuthentication led to a method vending a GTMFetcherAuthorizationProtocol conforming instance to not be visible in Swift. Fixes #106.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDAuthentication.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDAuthentication.h
@@ -16,7 +16,14 @@
 
 #import <Foundation/Foundation.h>
 
-@protocol GTMFetcherAuthorizationProtocol;
+// We have to import GTMAppAuth because forward declaring the protocol does
+// not generate the `fetcherAuthorizer` method below for Swift.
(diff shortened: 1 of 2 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDAuthentication.h
```diff
@@ -16,7 +16,12 @@
 
 #import <Foundation/Foundation.h>
 
-@protocol GTMFetcherAuthorizationProtocol;
+#ifdef SWIFT_PACKAGE
```
petea, line 19: Might add a comment explaining why we're not forward declaring.

petea (approved):

PR #111, opened 2022-03-29
Modify sample app `DaysUntilBirthday` to be multiplatform

Alex-4-Git:


petea (changes requested):
A few thoughts...
- Consider renaming the "Sources" folder to "Shared" to align with Xcode's multiplatform app template.
- Likewise, target names could be "DaysUntilBirthday (iOS)" and "DaysUntilBirthday (macOS)".
- Move the "GoogleSignIn-iOS" local package resource out of "Sources" and into the top level of the project resource tree.

petea:
Are you still planning on relocating /Shared/Views/[iOS, macOS] to /[iOS, macOS]?

petea (approved):

PR #113, opened 2022-04-05
Report execution environment via `gidenv`

petea:


paulb777 (approved):

GoogleSignIn/Sources/GIDSignInPreferences.m
```diff
@@ -39,6 +50,36 @@
   return [NSString stringWithFormat:@"gid-%@", @STR(GID_SDK_VERSION)];
 }
 
+// Get the current Apple execution environment.
+NSString* GIDEnvironment(void) {
```
mdmathias, line 54: Is this worth a unit test?

mdmathias (commented):
Left a question about unit testing the environment method.

mdmathias (approved):

GoogleSignIn/Sources/GIDSignInPreferences.m
petea: Yes it is.  Not sure how to cover all the scenarios here, but we can check a few of them.

PR #118, opened 2022-04-13
Add defaults for easier button instantiation

mdmathias:
Adds default values to the initializers for `GoogleSignInButtonViewModel` and `GoogleSignInButton` to facilitate easier instantiation.

### DIFF
--- a/GoogleSignInSwift/Sources/GoogleSignInButton.swift
+++ b/GoogleSignInSwift/Sources/GoogleSignInButton.swift
@@ -28,10 +28,12 @@ public struct GoogleSignInButton: View {
 
   /// Creates an instance of the Google Sign-In button for use in SwiftUI
   /// - parameter viewModel: An instance of `GoogleSignInButtonViewModel`
-  /// containing information on the button's scheme, style, and state.
+  /// containing information on the button's scheme, style, and state. Defaults to
+  /// `GoogleSignInButtonViewModel` with its standard defaults.
   /// - parameter action: A closure to use as the button's action upon press.
+  /// - seealso: Refer to `GoogleSignInButtonViewModel.swift` for its defaults.
   public init(
-    viewModel: GoogleSignInButtonViewModel,
+    viewModel: GoogleSignInButtonViewModel = GoogleSignInButtonViewModel(),
     action: @escaping () -> Void
   ) {
     self.viewModel = viewModel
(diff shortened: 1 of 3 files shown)


petea:
For setting style and scheme properties, should we consider using a pattern similar to that of the [Button](https://developer.apple.com/documentation/swiftui/button) view with its [buttonStyle](https://developer.apple.com/documentation/swiftui/view/buttonstyle(_:)-7qx1) instance method?

mdmathias:
@petea can you elaborate on your question regarding `buttonStyle`? We are currently utilizing a [custom `ButtonStyle` defined here](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignInSwift/Sources/GoogleSignInButtonStyling.swift#L252).

This is [set on the button here(https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignInSwift/Sources/GoogleSignInButton.swift#L80).

GoogleSignInSwift/Sources/GoogleSignInButton.swift
```diff
@@ -22,23 +22,49 @@ import CoreGraphics
...
   ) {
     self.viewModel = viewModel
     self.action = action
     self.fontLoaded = Font.loadCGFont()
   }
 
+  /// A convenience initializer to create a Google Sign-In button in SwiftUI
+  /// with scheme, style, and state.
+  /// - parameter scheme: The `GoogleSignInButtonColorScheme` to use. Defaults
+  /// to `.light`.
```
petea, line 48: Indent these wrapped comment lines?

GoogleSignInSwift/Sources/GoogleSignInButtonViewModel.swift
```diff
@@ -31,14 +31,16 @@ public class GoogleSignInButtonViewModel: ObservableObject {
   }
 
   /// Creates instances of the SwiftUI sign-in button.
-  /// - parameter scheme: An instance of `GoogleSignInButtonColorScheme`.
-  /// - parameter style: An instance of `GoogleSignInButtonStyle`.
+  /// - parameter scheme: An instance of `GoogleSignInButtonColorScheme`. Defaults to
+  /// `.light`.
```
petea, line 35: Same as above.

petea (approved):
Much easier to configure now.

Separately, at some point, we might want to think about adding a button configurator view to the Swift sample similar to the Objc sample.  This would help with exploration and testing.

mdmathias:
Added a [new issue](https://github.com/google/GoogleSignIn-iOS/issues/125) to track the addition of some way to update the button's appearance in the sample app.

PR #123, opened 2022-04-19
Set up example app for UI automation testing

mdmathias:
This change adds a new sample app set up only for the purpose of UI Automation testing the SwiftUI Google Sign-in button.


.github/workflows/tests.yml
```diff
@@ -61,3 +61,35 @@ jobs:
...
+    defaults:
+      run:
+        working-directory: Samples/Swift/DaysUntilBirthday
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Create xcconfig file
+      shell: bash
+      env:
+        EMAIL_SECRET : ${{ secrets.EMAIL_SECRET }}
```
petea, line 76: Remind me what type of Gaia account we're currently using with these secrets?

.github/workflows/tests.yml
```diff
@@ -61,3 +61,35 @@ jobs:
...
+        PASSWORD_SECRET : ${{ secrets.PASSWORD_SECRET }}
+      run: |
+        echo EMAIL_SECRET = $EMAIL_SECRET$'\n'PASSWORD_SECRET = $PASSWORD_SECRET >> DaysUntilBirthdayUITests\(iOS\)/Credentials.xcconfig
```
petea, line 79: Did you explore a means of injecting the credentials without writing them to disk, such as using compiler flags?

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS_.swift
```diff
@@ -0,0 +1,243 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import XCTest
+
+class DaysUntilBirthdayUITests_iOS_: XCTestCase {
```
petea, line 19: Why the trailing underscore in the class name?

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
```diff
@@ -57,9 +57,7 @@ struct UserProfileView: View {
         .toolbar {
           ToolbarItemGroup(placement: .navigationBarTrailing) {
             Button(NSLocalizedString("Disconnect", comment: "Disconnect button"), action: disconnect)
-              .accessibilityLabel(Text("Disconnect scope button."))
```
petea, line 60: What were these here for?

petea:
Could you sync this branch with main?  Looks like there are a few conflicts to resolve.

.github/workflows/tests.yml
mdmathias: It's an owned test account that I have.

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
mdmathias: I was originally overzealous with adding `accessibilityLabel(_:)` before. It's useful for labeling a view without text, but since this is a button with a clear label, we don't need it.

[Documentation](https://developer.apple.com/documentation/swiftui/view/accessibilitylabel(_:)-1d7jv)

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS_.swift
mdmathias: Good catch. I think this was automatically generated by adding the UI testing target with the `(iOS)` at the end of its name. I've fixed this.

.github/workflows/tests.yml
mdmathias: I have tested adding an environment variable to the iOS scheme's Test action. I _think_ we could pass the secrets in there like so (I have gotten this to work locally):

```
xcodebuild \
          -project Samples/Swift/DaysUntilBirthday/DaysUntilBirthday.xcodeproj \
          test-without-building \
          -scheme DaysUntilBirthday\(iOS\) \
          -sdk iphonesimulator \
          -destination 'platform=iOS Simulator,name=iPhone 11' \
          SECRET_EMAIL=$EMAIL_SECRET
```

However, two things stand out to me: 

1. As things stand now, the output for `xcodebuild` print the value of the secrets. Per a local test, it'll look something like this:

```
Build settings from command line:
    FOO_BAR = testValue
    SDKROOT = iphonesimulator15.4
```

I'm guessing that we can have less logged here, but then perhaps that will make debugging more difficult on GitHub.

2. Testing locally will require using `xcodebuild` from the command line, or setting up environment variables with hardcoded values for the iOS scheme's Test action. This will either mean not using Xcode directly for testing locally, or require a bit more setup (IMO).

mdmathias: Resolving as I was able to use environment variables passed in through `xcodebuild`.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/xcshareddata/xcschemes/DaysUntilBirthday(macOS).xcscheme
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
petea, line 1: Should we be adding a macOS scheme for the Swift project as well?  It looks like we're only adding an iOS scheme there.

mdmathias: Added.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/xcshareddata/xcschemes/DaysUntilBirthday(iOS).xcscheme
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
petea, line 1: For the two Pod project schemes, could you include a space between "DaysUntilBirthday" and "(iOS)" or "(macOS)"?  This would be consistent with the SPM project scheme names.

petea (approved):

PR #149, opened 2022-05-20
Add workflow notifying team chat about pull requests

mdmathias:


.github/workflows/notification.yml
```diff
@@ -0,0 +1,95 @@
+name: GSI Chatroom PR Notification
+
+on:
+  pull_request:
+    types: [opened, reopened]
```
Alex-4-Git, line 5: Does it make sense to use `review_requested` here? When a PR opens it is possible not ready yet.

.github/workflows/notification.yml
```diff
@@ -0,0 +1,95 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Pull Request Details
+      run: |
+        echo "Pull Request: ${{ github.event.pull_request.title }}"
+        echo "Author: ${{ github.event.pull_request.user.login }}"
+
+    - name: Google Chat Notification
+      run: |
+        curl --location --request POST '${{ secrets.WEBHOOK_URL }}' \
```
Alex-4-Git, line 22: Where are the secrets stored?

Alex-4-Git (commented):
Well done, Matt.

.github/workflows/notification.yml
mdmathias: They are stored on the [repo environment](https://docs.github.com/en/actions/security-guides/encrypted-secrets).

.github/workflows/notification.yml
mdmathias: Good question and point. I think it's nice to have a notification on `opened` as well. I've added a couple more that I think will be useful. Here is the [full list](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-34); take a look and let me know if you think others will be good to add.

mdmathias: We call always modify this list if things get to noisy in the channel.

petea (approved):
Let's give it a shot :)

PR #157, opened 2022-05-26
Prevent build errors for GoogleSignInSwift when using SPM

petea:
Because SPM does not support target-specific platform min versions, our GoogleSignInSwift target is built for both arm64 and armv7.  The armv7 stage of the build fails with missing SwiftUI symbol errors.  The armv7 slice is unnecessary as this target can only be used on iOS 13+.  This change works around the problem by preventing the armv7 build stage from doing any work.

### DIFF
--- a/GoogleSignInSwift/Sources/GoogleSignInButton.swift
+++ b/GoogleSignInSwift/Sources/GoogleSignInButton.swift
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#if !arch(arm)
+
 import SwiftUI
 import CoreGraphics
 
@@ -191,3 +193,5 @@ private extension Font {
     return CTFontManagerRegisterGraphicsFont(newFont, nil)
   }
 }
+
+#endif // !arch(arm)
--- a/GoogleSignInSwift/Sources/GoogleSignInButtonBundleExtensions.swift
+++ b/GoogleSignInSwift/Sources/GoogleSignInButtonBundleExtensions.swift
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#if !arch(arm)
+
 import Foundation
 
 // MARK: - Bundle Extensions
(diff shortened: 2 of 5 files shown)


paulb777 (approved):
Kind of ugly, but I suppose it works.

We avoided the arm7/SPM issue in Firebase by requiring iOS 11 for SPM since SPM users don't have the legacy requirements of CocoaPods issues and haven't seen any complaints. https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L26

petea:
> Kind of ugly, but I suppose it works.

Unfortunately `#if canImport(SwiftUI)` succeeds in the armv7 scenario here, so that wasn't an option.  I'd much prefer a SPM solution but I'm not seeing one short of moving the target into a separate repo and package.

gmogames:
@petea do you think this will get merged and the 6.2.2 released today? This is breaking our ci pipelines for some of our environments and would be very good to have it merged before the holiday next week. Thank you!

petea:
@gmogames we're aiming to release 6.2.2 in the next few hours.

gmogames:
Thank you!

petea:
@gmogames note that, as long as you're not using the new SwiftUI button, you can also remove the GoogleSignInSwift library from your project as SPM may have added it automatically.

macintosh-HD:
Unfortunately I have to report that everything works fine for me as long as the build configuration is set to "Debug". As soon as I switch to "Release" I get the same errors again. Is there something I'm overlooking on my side? Or can you recreate this behaviour?

petea:
@macintosh-HD I'm not able to recreate this, having looked briefly.  Building with a release configuration using Archive or otherwise is working as expected for me.  Can you provide detailed repro steps and tell me more about what you're building?

petea:
@macintosh-HD thanks for bringing this to my attention.  I've now been able to repo and a fix is on the way.

PR #175, opened 2022-07-01
Disable SwiftUI button test

mdmathias:


.github/workflows/tests.yml
```diff
@@ -10,6 +10,8 @@ on:
   workflow_dispatch:
 
 jobs:
+  swift-button-functional-test:
```
petea, line 13: Shouldn't this be part of the existing job definition below?

petea:
Looks like you've got a yaml syntax error in tests.yml.

petea (approved):

PR #177, opened 2022-07-05
Create class GIDUserAuth

Alex-4-Git:
Add class GIDUserAuth to represent the outcome of a successful signIn or addScopes flow.

### DIFF
--- a/GoogleSignIn/Sources/GIDUserAuth_Private.h
+++ b/GoogleSignIn/Sources/GIDUserAuth_Private.h
@@ -22,6 +22,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface GIDUserAuth ()
 
 // Private initializer for |GIDUserAuth|.
+// @param user The current GIDGoogleUser.
+// @param severAuthCode The one-time authorization code for backend to exchange
+//     access and refresh tokens.
 - (instancetype)initWithGoogleUser:(GIDGoogleUser *)user
                     serverAuthCode:(nullable NSString *)serverAuthCode;
 
--- a/GoogleSignIn/Sources/GIDUserAuth_Private.h
+++ b/GoogleSignIn/Sources/GIDUserAuth_Private.h
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDUserAuth.h
```diff
@@ -0,0 +1,38 @@
...
+/// A helper object that contains the outcome of a successful signIn or addScopes flow.
+@interface GIDUserAuth : NSObject
+
+/// The updated `GIDGoogleUser` instance for the user who just completed the flow.
+@property(nonatomic, readonly) GIDGoogleUser *user;
+
+/// An OAuth2 authorization code for the home server.
+@property(nonatomic, readonly, nullable) NSString *serverAuthCode;
+
++ (instancetype)new NS_UNAVAILABLE;
```
mdmathias, line 32: Is it worth documenting here how this type is supposed to be created?

GoogleSignIn/Sources/GIDUserAuth_Private.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDUserAuth.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+// Private |GIDUserAuth| methods that are used in this SDK.
+@interface GIDUserAuth ()
+
+// Private initializer for |GIDUserAuth|.
```
mdmathias, line 24: This initializer is scoped internally to `GoogleSignIn`, right? If so, I'd suggest writing a doc comment for this initializer since it is part of `GIDUserAuth`'s API that will visible and used there.

mdmathias (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDUserAuth.h
mdmathias: I don't think `/// Unsupported.` adds much. It'd be more helpful to have something like (for this and `init` below):

`Unavailable. See the internally scoped `-[GIDUserAuth_Private initWithGoogleUser:serverAuthCode:]` to create an instance.

mdmathias: Fair enough. I don't think `/// Unsupported.` is useful. Feel free to remove.

PR #182, opened 2022-07-17
Make asynchronous methods callable via async/await

### LINKED ISSUE #91: Enable automatic async/await API generation
The GSI completion handlers use `callback` as the parameter selector for specifying completion handlers. 

`callback` is not one of the selectors included for automatic async method generation. See https://developer.apple.com/documentation/swift/calling_objective-c_apis_asynchronously

If I manually change the selector in the GIDSignIn.h public header to `completion`, Xcode will auto-generate the API:

Screen Shot 2022-01-14 at 2 54 16 PM

peterfriese:
This fixes #91

@mdmathias @petea - can you provide guidance on formatting / indentation? I will also look into updating the sample as part of this PR.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
@@ -201,7 +201,7 @@ typedef void (^GIDDisconnectCallback)(NSError *_Nullable error);
 ///     called asynchronously on the main queue.
 - (void)signInWithConfiguration:(GIDConfiguration *)configuration
                presentingWindow:(NSWindow *)presentingWindow
-                       callback:(nullable GIDSignInCallback)callback;
+                       callback:(nullable GIDSignInCallback)callback
     __attribute__((swift_async_name("signIn(with:presenting:)")));
 
 /// Starts an interactive sign-in flow on macOS using the provided configuration and a login hint.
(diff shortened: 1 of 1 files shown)


mdmathias:
Hi @peterfriese thanks for the PR! It looks like it fails some checks during `pod lib lint` on macOS 10.15.

I was thinking that we'd address #91 by updating the block parameter names to be `completion`.

peterfriese:
Let me look into the lint checks.

I was thinking about renaming the block parameter to `completion` as well at first, but was worried that this would be a breaking change for the API. What are your thoughts about this?

mdmathias:
You're correct - changing the block parameter name to `completion` would break the API and require a major version release. Does adding the `swift_async_name` attribute not break the API? I guess it just adds an async version of the method and doesn't replace the pre-existing version. I'm not as familiar with with this attribute...

In any case, if `swift_async_name` doesn't break the API, and only adds async methods, then I think it's worth figuring out what is causing issues in the linting.

mdmathias:
Another consideration is that we are due to have a major release coming up soon anyway. Perhaps it's okay to change this API? What do you think @petea?

mdmathias:
I did a little informal test and it looks like adding the `swift_async_name` attribute just adds an `async` method, but leaves the existing API unchanged.

Without the attribute:
Screen Shot 2022-07-20 at 2 59 02 PM

With attribute:
Screen Shot 2022-07-20 at 3 04 45 PM

petea:
@peterfriese thanks for bringing `swift_async_name` to our attention.  This could allow us to ship Swift Concurrency support in a non-breaking fashion.  However, we are planning a major release toward the end of Q3 and it may be preferable to take that opportunity to rename callback -> completion for automatic Swift async method generation as well as slightly more idiomatic method naming in general.  In either case, `swift_async_name` might be useful for `doWithFreshTokens:` in `GIDAuthentication`, as this method doesn't even begin to conform to the async naming conventions.

petea:
Addressed with #187 and will go out in our next major release.

PR #188, opened 2022-07-27
Update macos runners to explicitly use 11 and 12

mdmathias:
macos-10.15 is deprecated, and will be fully unsupported on [August 30](https://github.com/actions/virtual-environments/issues/5583).

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, macos-10.15]
+        os: [macos-11, macos-12]
   
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/builds.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, macos-10.15]
+        os: [macos-latest, macos-11]
```
petea, line 13: The `macos-latest` virtual environment is currently [defined](https://github.com/actions/virtual-environments#available-environments) as macOS 11.  Maybe we should be more direct with `[macos-11, macos-12]`?

mdmathias: Good suggestion.

petea (approved):

PR #189, opened 2022-07-28
Add Three token properties in GIDGoogleUser API

Alex-4-Git:
Move the token related properties from GIDAuthentication.h to GIDGoogleUser.h. 
In the previous PR we created the new class GIDToken. We can use this new type to add three tokens: access token, refresh token and ID token.

### DIFF
--- a/GoogleSignIn/Sources/GIDGoogleUser.m
+++ b/GoogleSignIn/Sources/GIDGoogleUser.m
@@ -20,6 +20,7 @@
 
 #import "GoogleSignIn/Sources/GIDAuthentication_Private.h"
 #import "GoogleSignIn/Sources/GIDProfileData_Private.h"
+#import "GoogleSignIn/Sources/GIDToken_Private.h"
 
 #ifdef SWIFT_PACKAGE
 @import AppAuth;
@@ -47,7 +48,7 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
 }
 
 - (nullable NSString *)userID {
-  NSString *idToken = [self idToken];
+  NSString *idToken = [self idTokenString];
   if (idToken) {
     OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idToken];
     if (idTokenDecoded && idTokenDecoded.subject) {
@@ -106,6 +107,7 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
   _authState = authState;
   _authentication = [[GIDAuthentication alloc] initWithAuthState:authState];
   _profile = profileData;
+  [self updateTokens];
 }
 
 #pragma mark - Helpers
@@ -115,18 +117,14 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
 }
 
 - (nullable NSString *)hostedDomain {
-  NSString *idToken = [self idToken];
-  if (idToken) {
-    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idToken];
+  NSString *idTokenString = [self idTokenString];
+  if (idTokenString) {
+    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idTokenString];
     if (idTokenDecoded && idTokenDecoded.claims[kHostedDomainIDTokenClaimKey]) {
       return [idTokenDecoded.claims[kHostedDomainIDTokenClaimKey] copy];
     }
   }
   return nil;
-}
-
-- (NSString *)idToken {
-  return _authState ? _authState.lastTokenResponse.idToken : nil;
 }
 
 - (nullable NSString *)serverClientID {
@@ -135,6 +133,40 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
 
 - (nullable NSString *)openIDRealm {
   return [_authState.lastTokenResponse.request.additionalParameters[kOpenIDRealmParameter] copy];
+}
+
+- (NSString *)idTokenString {
+  return _authState.lastTokenResponse.idToken;
+}
+
+- (nullable NSDate *)idTokenExpirationDate {
+  return [[[OIDIDToken alloc] initWithIDTokenString:self.idTokenString] expiresAt];
+}
+
+- (NSString *)accessTokenString {
+  return _authState.lastTokenResponse.accessToken;
+}
+
+- (NSDate *)accessTokenExpirationDate {
+  return _authState.lastTokenResponse.accessTokenExpirationDate;
+}
+
+- (NSString *)refreshTokenString {
+  return _authState.refreshToken;
+}
+
+- (void)updateTokens {
+  _accessToken = [[GIDToken alloc] initWithTokenString:[self accessTokenString]
+                                        expirationDate:[self accessTokenExpirationDate]];
+  _refreshToken = [[GIDToken alloc] initWithTokenString:[self refreshTokenString]
+                                         expirationDate:nil];
+  NSString *idTokenString = [self idTokenString];
+  if (idTokenString) {
+    _idToken = [[GIDToken alloc] initWithTokenString:[self idTokenString]
+                                      expirationDate:[self idTokenExpirationDate]];
+  } else {
+    _idToken = nil;
+  }
 }
 
 #pragma mark - NSSecureCoding
@@ -155,6 +187,7 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
       _authState = authentication.authState;
     }
     _authentication = [[GIDAuthentication alloc] initWithAuthState:_authState];
+    [self updateTokens];
   }
   return self;
 }
--- a/GoogleSignIn/Sources/GIDGoogleUser.m
+++ b/GoogleSignIn/Sources/GIDGoogleUser.m
@@ -48,7 +48,7 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
 }
 
 - (nullable NSString *)userID {
-  NSString *idToken = [self idTokenString];
+  NSString *idToken = [self lastTokenResponseToken];
   if (idToken) {
     OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idToken];
     if (idTokenDecoded && idTokenDecoded.subject) {
@@ -117,9 +117,9 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
 }
 
 - (nullable NSString *)hostedDomain {
-  NSString *idTokenString = [self idTokenString];
-  if (idTokenString) {
-    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idTokenString];
+  NSString *lastTokenResponseToken = [self lastTokenResponseToken];
+  if (lastTokenResponseToken) {
+    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:lastTokenResponseToken];
     if (idTokenDecoded && idTokenDecoded.claims[kHostedDomainIDTokenClaimKey]) {
       return [idTokenDecoded.claims[kHostedDomainIDTokenClaimKey] copy];
     }
@@ -135,15 +135,15 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
   return [_authState.lastTokenResponse.request.additionalParameters[kOpenIDRealmParameter] copy];
 }
 
-- (NSString *)idTokenString {
+- (NSString *)lastTokenResponseToken {
   return _authState.lastTokenResponse.idToken;
 }
 
 - (nullable NSDate *)idTokenExpirationDate {
-  return [[[OIDIDToken alloc] initWithIDTokenString:self.idTokenString] expiresAt];
+  return [[[OIDIDToken alloc] initWithIDTokenString:[self lastTokenResponseToken]] expiresAt];
 }
 
-- (NSString *)accessTokenString {
+- (NSString *)lastTokenResponseAccessToken {
   return _authState.lastTokenResponse.accessToken;
 }
 
@@ -151,18 +151,14 @@ static NSString *const kOpenIDRealmParameter = @"openid.realm";
   return _authState.lastTokenResponse.accessTokenExpirationDate;
 }
 
-- (NSString *)refreshTokenString {
-  return _authState.refreshToken;
-}
-
 - (void)updateTokens {
-  _accessToken = [[GIDToken alloc] initWithTokenString:[self accessTokenString]
+  _accessToken = [[GIDToken alloc] initWithTokenString:[self lastTokenResponseAccessToken]
                                         expirationDate:[self accessTokenExpirationDate]];
-  _refreshToken = [[GIDToken alloc] initWithTokenString:[self refreshTokenString]
+  _refreshToken = [[GIDToken alloc] initWithTokenString:_authState.refreshToken
                                          expirationDate:nil];
-  NSString *idTokenString = [self idTokenString];
-  if (idTokenString) {
-    _idToken = [[GIDToken alloc] initWithTokenString:[self idTokenString]
+  NSString *lastTokenResponseToken = [self lastTokenResponseToken];
+  if (lastTokenResponseToken) {
+    _idToken = [[GIDToken alloc] initWithTokenString:lastTokenResponseToken
                                       expirationDate:[self idTokenExpirationDate]];
   } else {
     _idToken = nil;
--- a/GoogleSignIn/Sources/GIDGoogleUser.m
+++ b/GoogleSignIn/Sources/GIDGoogleUser.m
@@ -45,9 +45,9 @@ NS_ASSUME_NONNULL_BEGIN
 @implementation GIDGoogleUser {
   OIDAuthState *_authState;
   GIDConfiguration *_cachedConfiguration;
-  GIDToken *_accessToken;
-  GIDToken *_refreshToken;
-  GIDToken *_idToken;
+  GIDToken *_cachedAccessToken;
+  GIDToken *_cachedRefreshToken;
+  GIDToken *_cachedIdToken;
 }
 
 - (nullable NSString *)userID {
@@ -103,36 +103,36 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (GIDToken *)accessToken {
   @synchronized(self) {
-    if (!_accessToken) {
-      _accessToken = [[GIDToken alloc] initWithTokenString:_authState.lastTokenResponse.accessToken
-                                            expirationDate:_authState.lastTokenResponse.
-                                                               accessTokenExpirationDate];
+    if (!_cachedAccessToken) {
+      _cachedAccessToken = [[GIDToken alloc] initWithTokenString:_authState.lastTokenResponse.accessToken
+                                                  expirationDate:_authState.lastTokenResponse.
+                                                                     accessTokenExpirationDate];
     }
   }
-  return _accessToken;
+  return _cachedAccessToken;
 }
 
 - (GIDToken *)refreshToken {
   @synchronized(self) {
-    if (!_refreshToken) {
-      _refreshToken = [[GIDToken alloc] initWithTokenString:_authState.refreshToken
-                                             expirationDate:nil];
+    if (!_cachedRefreshToken) {
+      _cachedRefreshToken = [[GIDToken alloc] initWithTokenString:_authState.refreshToken
+                                                   expirationDate:nil];
     }
   }
-  return _refreshToken;
+  return _cachedRefreshToken;
 }
 
 - (nullable GIDToken *)idToken {
   @synchronized(self) {
     NSString *idTokenString = _authState.lastTokenResponse.idToken;
-    if (!_idToken && idTokenString) {
+    if (!_cachedIdToken && idTokenString) {
       NSDate *idTokenExpirationDate = [[[OIDIDToken alloc]
                                         initWithIDTokenString:idTokenString] expiresAt];
-      _idToken = [[GIDToken alloc] initWithTokenString:idTokenString
-                                        expirationDate:idTokenExpirationDate];
+      _cachedIdToken = [[GIDToken alloc] initWithTokenString:idTokenString
+                                              expirationDate:idTokenExpirationDate];
     }
   }
-  return _idToken;
+  return _cachedIdToken;
 }
 
 #pragma mark - Private Methods
@@ -154,9 +154,9 @@ NS_ASSUME_NONNULL_BEGIN
     _profile = profileData;
     
     // These three tokens will be generated in the getter and cached .
-    _accessToken = nil;
-    _refreshToken = nil;
-    _idToken = nil;
+    _cachedAccessToken = nil;
+    _cachedRefreshToken = nil;
+    _cachedIdToken = nil;
   }
 }
 
--- a/GoogleSignIn/Sources/GIDGoogleUser.m
+++ b/GoogleSignIn/Sources/GIDGoogleUser.m
@@ -28,8 +28,6 @@
 #import <AppAuth/AppAuth.h>
 #endif
 
-NS_ASSUME_NONNULL_BEGIN
-
 // The ID Token claim key for the hosted domain value.
 static NSString *const kHostedDomainIDTokenClaimKey = @"hd";
 
@@ -42,10 +40,16 @@ static NSString *const kAuthState = @"authState";
 static NSString *const kAudienceParameter = @"audience";
 static NSString *const kOpenIDRealmParameter = @"openid.realm";
 
+NS_ASSUME_NONNULL_BEGIN
+
 @implementation GIDGoogleUser {
   OIDAuthState *_authState;
   GIDConfiguration *_cachedConfiguration;
 }
+
+@synthesize accessToken = _accessToken;
+@synthesize refreshToken = _refreshToken;
+@synthesize idToken = _idToken;
 
 - (nullable NSString *)userID {
   NSString *idTokenString = self.idToken.tokenString;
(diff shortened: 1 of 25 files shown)


mdmathias:
It might be beneficial in the future to add some more context in your PR description. Off the top of my head, I'm not sure why we want to add the these three token properties. I can go looking for context to help, but it'd be nice to have it here too. Moreover, I think we are all much less likely to remember, or be able to find, this context if we return to this PR in the future.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,40 @@ - (nullable NSString *)openIDRealm {
   return [_authState.lastTokenResponse.request.additionalParameters[kOpenIDRealmParameter] copy];
 }
 
+- (NSString *)idTokenString {
```
mdmathias, line 138: In general, I think it's useful to avoid using type information in the name of a method, property, or variable. I typically do so only to avoid a collision with something other name. Another use case that I think makes sense is if your method provides a `String` representation of another type.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +55,37 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+  }
+  return [self isEqualToGIDToken:(GIDToken *)object];
+}
+
+- (BOOL)isEqualToGIDToken:(GIDToken *)other {
+  return [_tokenString isEqual:other.tokenString] &&
+      [self isTheSameDateWithDate1:_expirationDate date2:other.expirationDate];
+}
+
+- (BOOL)isTheSameDateWithDate1:(NSDate *)date1
```
mdmathias, line 75: The parameter names shouldn't be numbered. Try to think of a more descriptive name.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +55,37 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+                         date2:(NSDate *)date2 {
+  // The date is nullable. Two `null` date is equal.
+  if (!date1 && !date2) {
```
mdmathias, line 78: Is this the purpose of the method? I'm not sure what the role of this method is given that you call through to `isEqualToDate:`. 

I don't understand why you want nil and nil to return an equal date.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
```diff
@@ -40,6 +41,17 @@ NS_ASSUME_NONNULL_BEGIN
 /// The configuration that was used to sign in this user.
 @property(nonatomic, readonly) GIDConfiguration *configuration;
 
+/// The OAuth2 access token to access Google services.
+@property(nonatomic, readonly) GIDToken *accessToken;
+
+/// The OAuth2 refresh token to exchange for new access tokens.
+@property(nonatomic, readonly) GIDToken *refreshToken;
+
+/// An OpenID Connect ID token that identifies the user. Send this token to your server to
```
mdmathias, line 50: We typically want doc comments to be a single sentence. You could consider adding a `note` for the second sentence. Look around for other examples in the repo and match that style.

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: Yes, this is to avoid the naming conflict with idToken which is a GIDToken type.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: This method is to facilitate `isEqualToGIDToken:`. If two tokens have the same token strings and they don't have an expiration date, they are equal.
Do you have any thought to make this clearer? I have not found the best way to compare two nullable properties.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
Alex-4-Git: This is the style in `GoogleSignIn`. We can discuss that and use another PR to change them altogether.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,40 @@ - (nullable NSString *)openIDRealm {
...
+  return _authState.lastTokenResponse.idToken;
+}
+
+- (nullable NSDate *)idTokenExpirationDate {
+  return [[[OIDIDToken alloc] initWithIDTokenString:self.idTokenString] expiresAt];
+}
+
+- (NSString *)accessTokenString {
```
mdmathias, line 146: This is another example where you append the type of the property to its name.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,40 @@ - (nullable NSString *)openIDRealm {
...
+  return _authState.lastTokenResponse.accessToken;
+}
+
+- (NSDate *)accessTokenExpirationDate {
+  return _authState.lastTokenResponse.accessTokenExpirationDate;
+}
+
+- (NSString *)refreshTokenString {
```
mdmathias, line 154: And here.

GoogleSignIn/Sources/GIDGoogleUser.m
mdmathias: Is there actually a naming conflict? It looks like you removed the getter for `idToken` and replaced it with `idTokenString`.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,40 @@ - (nullable NSString *)openIDRealm {
   return [_authState.lastTokenResponse.request.additionalParameters[kOpenIDRealmParameter] copy];
 }
 
+- (NSString *)idTokenString {
+  return _authState.lastTokenResponse.idToken;
+}
+
+- (nullable NSDate *)idTokenExpirationDate {
```
mdmathias, line 142: Personally, I don't think you need `Date` at the end of this. 'Expiration` already connotes time.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,40 @@ - (nullable NSString *)openIDRealm {
...
+  return [[[OIDIDToken alloc] initWithIDTokenString:self.idTokenString] expiresAt];
+}
+
+- (NSString *)accessTokenString {
+  return _authState.lastTokenResponse.accessToken;
+}
+
+- (NSDate *)accessTokenExpirationDate {
```
mdmathias, line 150: Same point regarding 'Expiration' and 'Date' in the name being redundant.

GoogleSignIn/Sources/GIDGoogleUser.m
mdmathias: As an example, consider `+[NSDate distantFuture]`. The word 'Future' already tells you to that the thing returned is related to time. The getter's signature tells you it is an `NSDate`. Under the naming you're using here, this method would look like, `+[NSDate distantFutureDate]`, which reads redundantly to me.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
mdmathias: We're already using the style that I mentioned. 

https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h#L43

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +55,37 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+    return YES;
+  }
+  if (!date1 || !date2) {
```
mdmathias, line 81: Is this check needed? If either is nil, then won't `-[NSDate isEqualToDate:]` return `NO`?

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +55,37 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+  if (self == object) {
+    return YES;
+  }
+  if (![object isKindOfClass:[GIDToken class]]) {
+    return NO;
+  }
+  return [self isEqualToGIDToken:(GIDToken *)object];
+}
+
+- (BOOL)isEqualToGIDToken:(GIDToken *)other {
```
mdmathias, line 70: I don't think you need to have the full type as the last phrase of this method. Consider `-[NSDate isEqualToDate:]` as opposed to `-[NSDate isEqualToNSDate:]`.

GoogleSignIn/Sources/GIDToken.m
mdmathias: I think you should clarify your comment on line 77, which is ambiguous in this context since you don't reference _why_ you're considering two nil dates as equal. A comment like this could clarify the issue: `// Nil dates considered equal so token equality check succeeds if token strings are the same and have no expiration.`

GoogleSignIn/Tests/Unit/GIDTokenTest.m
```diff
@@ -44,8 +54,7 @@ - (void)testCoding {
     GIDToken *newToken = [NSKeyedUnarchiver unarchivedObjectOfClass:[GIDToken class]
```
mdmathias, line 54: These parameters are not aligned.

GoogleSignIn/Tests/Unit/GIDTokenTest.m
```diff
@@ -36,6 +36,16 @@ - (void)testInitializer {
   XCTAssertEqualObjects(token.tokenString, tokenString);
   XCTAssertEqualObjects(token.expirationDate, _date);
 }
+
```
mdmathias, line 39: You should add a test documenting how two tokens are equal if their token strings are equal but don't have expirations. You should also add a test showing that two tokens with equal strings, but one with an expiration and another without an expiration, are not equal.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: That's a good comment!

GoogleSignIn/Sources/GIDGoogleUser.m
mdmathias: I see the property for `idToken` now. Sorry for missing it. :) 

Given that this getter works with `_authState.lastTokenResponse.idToken`, I'd suggest not naming it `idTokenString` (i.e., you're not going through the property `idToken` on `GIDGoogleUser`). Maybe rename this to `lastTokenResponseIdToken` or something like that.

This suggestion applies here and throughout.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: Definitely don't need the full type.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -25,7 +25,7 @@
 @implementation GIDToken
```
petea, line 25: Need to use `NS_ASSUME_NONNULL_BEGIN` / `NS_ASSUME_NONNULL_END` for this implementation.

petea (changes requested):

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,39 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+
+- (BOOL)isEqualToToken:(GIDToken *)other {
+  return [_tokenString isEqual:other.tokenString] &&
+      [self isTheSameDate:_expirationDate with:other.expirationDate];
+}
+
+- (BOOL)isTheSameDate:(NSDate *)date1
+                 with:(NSDate *)date2 {
+  
+  if (date1 || date2) {
```
petea, line 80: Might be more intuitive to do something like:

```
if (date1 and date2 are nil) {
  return YES;
}
return [date1 isEqualToDate:date2];
```

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,39 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+  if (self == object) {
+    return YES;
+  }
+  if (![object isKindOfClass:[GIDToken class]]) {
+    return NO;
+  }
+  return [self isEqualToToken:(GIDToken *)object];
+}
+
+- (BOOL)isEqualToToken:(GIDToken *)other {
```
petea, line 72: We should declare this method in the public interface.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,39 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
   [encoder encodeObject:_expirationDate forKey:kExpirationDateKey];
 }
 
+#pragma mark - isEqual
+
+- (BOOL)isEqual:(id)object {
+  if (self == object) {
+    return YES;
+  }
+  if (![object isKindOfClass:[GIDToken class]]) {
```
petea, line 66: For clarity, you might take the nil handling out of this line and add it in a new condition at the top of the method:

```
if (object == nil) {
  return NO;
}
```

petea (changes requested):

mdmathias (approved):

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,39 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
   [encoder encodeObject:_expirationDate forKey:kExpirationDateKey];
 }
 
+#pragma mark - isEqual
+
+- (BOOL)isEqual:(id)object {
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,39 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+      [self isTheSameDate:_expirationDate with:other.expirationDate];
+}
+
+- (BOOL)isTheSameDate:(NSDate *)date1
+                 with:(NSDate *)date2 {
+  
+  if (date1 || date2) {
+    return [date1 isEqualToDate:date2];
+  } else {
+  // The date is nullable in GIDToken. Two `nil` dates are considered equal so
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: if (!date1 && !date2) is not working because of the short circuit. Checking at least one is not nil is easier.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDToken.m
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +135,36 @@ - (nullable NSString *)openIDRealm {
...
+
+- (NSString *)lastTokenResponseAccessToken {
+  return _authState.lastTokenResponse.accessToken;
+}
+
+- (NSDate *)accessTokenExpirationDate {
+  return _authState.lastTokenResponse.accessTokenExpirationDate;
+}
+
+- (void)updateTokens {
```
petea, line 154: The three new token properties are all derived from `_authState`.  Should we not use cached getters for them as we've done for `configuration`?

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -115,20 +117,16 @@ - (NSString *)clientID {
 }
 
 - (nullable NSString *)hostedDomain {
-  NSString *idToken = [self idToken];
-  if (idToken) {
-    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idToken];
+  NSString *lastTokenResponseToken = [self lastTokenResponseToken];
```
petea, line 120: Would probably be clearer to use `self.idToken.tokenString` here.

petea (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDToken.h
```diff
@@ -27,6 +27,11 @@ NS_ASSUME_NONNULL_BEGIN
 /// The estimated expiration date of the token.
 @property(nonatomic, readonly, nullable) NSDate *expirationDate;
 
+/// Check if current token is equal to another one.
+///
+/// @param other - Another token to compare.
+- (BOOL)isEqualToToken:(GIDToken *)other;
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDToken.m
```diff
@@ -55,4 +57,40 @@ - (void)encodeWithCoder:(NSCoder *)encoder {
...
+}
+
+- (BOOL)isEqualToToken:(GIDToken *)other {
+  return [_tokenString isEqual:other.tokenString] &&
+      [self isTheSameDate:_expirationDate with:other.expirationDate];
+}
+
+// The date is nullable in GIDToken. Two `nil` dates are considered equal so
+// token equality check succeeds if token strings are equal and have no expiration.
+- (BOOL)isTheSameDate:(NSDate *)date1
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Tests/Unit/GIDTokenTest.m
```diff
@@ -36,16 +37,46 @@ - (void)testInitializer {
...
+- (void)testTokensWithSameTokenStringAndExpirationDateAreEqual {
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: After consideration I removed this method and updated tokens in the method `updateAuthState:profileData:`

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: Good idea. It is clearer.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -137,6 +146,22 @@ - (nullable NSString *)openIDRealm {
   return [_authState.lastTokenResponse.request.additionalParameters[kOpenIDRealmParameter] copy];
 }
 
+- (NSString *)lastTokenResponseToken {
```
petea, line 149: The naming for these four helper methods should be consistent and clearly communicate what they are doing.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -106,6 +108,17 @@ - (void)updateAuthState:(OIDAuthState *)authState
   _authState = authState;
   _authentication = [[GIDAuthentication alloc] initWithAuthState:authState];
   _profile = profileData;
+  _accessToken = [[GIDToken alloc] initWithTokenString:[self lastTokenResponseAccessToken]
```
petea, line 111: When initializing instances of `GIDToken` or `GIDConfiguration` we should should `copy` the values.  We're currently doing this in some places and not in others.

petea (changes requested):

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: Done.

petea (changes requested):

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -41,15 +40,22 @@
...
 static NSString *const kOpenIDRealmParameter = @"openid.realm";
 
+NS_ASSUME_NONNULL_BEGIN
+
 @implementation GIDGoogleUser {
   OIDAuthState *_authState;
   GIDConfiguration *_cachedConfiguration;
 }
 
+@synthesize accessToken = _accessToken;
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -90,6 +96,40 @@ - (GIDConfiguration *)configuration {
   return _cachedConfiguration;
 }
 
+- (GIDToken *)accessToken {
+  @synchronized(self) {
+    if (!_accessToken) {
+      _accessToken = [[GIDToken alloc] initWithTokenString:[_authState.lastTokenResponse.accessToken copy]
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,32 +143,35 @@ - (instancetype)initWithAuthState:(OIDAuthState *)authState
...
+    _profile = profileData;
+    
+    // These three tokens will be generated in the getter and cached .
+    _accessToken = nil;
+    _refreshToken = nil;
+    _idToken = nil;
+  }
 }
 
 #pragma mark - Helpers
```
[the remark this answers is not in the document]
Alex-4-Git: Only keep the helper `hostedDomain` containing more logic.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -90,6 +96,40 @@ - (GIDConfiguration *)configuration {
...
+      _refreshToken = [[GIDToken alloc] initWithTokenString:[_authState.refreshToken copy]
+                                             expirationDate:nil];
+    }
+  }
+  return _refreshToken;
+}
+
+- (nullable GIDToken *)idToken {
+  @synchronized(self) {
+    NSString *idTokenString = [_authState.lastTokenResponse.idToken copy];
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -90,6 +96,40 @@ - (GIDConfiguration *)configuration {
...
+      NSDate *idTokenExpirationDate = [[[OIDIDToken alloc]
+                                        initWithIDTokenString:idTokenString] expiresAt];
+      _idToken = [[GIDToken alloc] initWithTokenString:idTokenString
+                                        expirationDate:idTokenExpirationDate];
+    }
+  }
+  return _idToken;
+}
+
+
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,32 +143,35 @@ - (instancetype)initWithAuthState:(OIDAuthState *)authState
...
+    _refreshToken = nil;
+    _idToken = nil;
+  }
 }
 
 #pragma mark - Helpers
 
 - (NSString *)clientID {
-  return _authState.lastAuthorizationResponse.request.clientID;
+  return [_authState.lastAuthorizationResponse.request.clientID copy];
```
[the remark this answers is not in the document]
Alex-4-Git: Removed this. GIDConfiguration has copy in its initializer.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,32 +143,35 @@ - (instancetype)initWithAuthState:(OIDAuthState *)authState
...
 
 - (nullable NSString *)hostedDomain {
-  NSString *idToken = [self idToken];
-  if (idToken) {
-    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idToken];
+  NSString *idTokenString = self.idToken.tokenString;
+  if (idTokenString) {
+    OIDIDToken *idTokenDecoded = [[OIDIDToken alloc] initWithIDTokenString:idTokenString];
     if (idTokenDecoded && idTokenDecoded.claims[kHostedDomainIDTokenClaimKey]) {
       return [idTokenDecoded.claims[kHostedDomainIDTokenClaimKey] copy];
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,32 +143,35 @@ - (instancetype)initWithAuthState:(OIDAuthState *)authState
...
   }
   return nil;
 }
 
-- (NSString *)idToken {
-  return _authState ? _authState.lastTokenResponse.idToken : nil;
-}
-
 - (nullable NSString *)serverClientID {
   return [_authState.lastTokenResponse.request.additionalParameters[kAudienceParameter] copy];
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,32 +143,35 @@ - (instancetype)initWithAuthState:(OIDAuthState *)authState
...
 }
```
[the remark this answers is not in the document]
Alex-4-Git: Done.

petea (changes requested):

petea (approved):

PR #190, opened 2022-07-29
Use async await API in DaysUntilBirthday Sample App

mdmathias:


petea (commented):
Should we update the `doWithFreshTokens` call in `BirthdayLoader` as well?

Alex-4-Git (approved):

PR #191, opened 2022-08-02
Add async and await support to DaysUntilBirthday Swift sample app

mdmathias:


petea:
Could you rebase this branch against `main` to exclude the already-merged API refactoring changes here?

mdmathias:
Yep! Sorry about that.

Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
```diff
@@ -8,5 +8,7 @@
 	<true/>
 	<key>com.apple.security.network.client</key>
 	<true/>
+	<key>keychain-access-groups</key>
```
petea, line 11: Did you mean to add this?

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -47,7 +47,35 @@ final class AuthenticationViewModel: ObservableObject {
 
   /// Signs the user in.
   func signIn() {
```
petea, line 49: Should this be async?

petea (changes requested):

Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
mdmathias: Yes, the sample won't run without it on macOS. I wonder if you see the same?

Samples/Swift/DaysUntilBirthday/DaysUntilBirthday.xcodeproj/project.pbxproj
```diff
@@ -656,6 +656,7 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_ASSET_PATHS = "\"macOS/Preview Content\"";
+				DEVELOPMENT_TEAM = "";
```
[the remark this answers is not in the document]
mdmathias: I'm just going to drop this since it snuck in here bc I was testing this on macOS, which requires signiing.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
mdmathias: Not as things stand now. The method currently uses a `Task` to update the sign in state on the main actor, which is required for updating observed objects that will impact view state (as this does). 

Also this method is given as the action in the `GoogleSignInButton`'s initializer. We may want to have another initializer for the button that does take an `async` method, but I think this is a separate conversation and project.

Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
petea: Are you signing the build on macOS?

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
petea: Sounds reasonable.  Just looking for ways to minimize the `Task` overhead.

Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
mdmathias: I've removed this change and created #192 to track this issue.

petea: I'm still seeing the change in this PR.

mdmathias: Hm, maybe it wasn't reverted before? In any case, it's no longer in this PR.

petea (approved):
Looks good.  Discussed offline and agreed that we would wait to merge into `main` until we're closer to shipping Swift Concurrency support.

Alex-4-Git (approved):

petea:
@mdmathias this is now merged and works as expected locally for me (including integration tests) but something seems to have broken integration tests on GitHub.  I'm wondering if it's Xcode 14 related as I'm still using 13 locally.

PR #193, opened 2022-08-05
Add keychain sharing entitlement to sign with dev certificate on macOS

### LINKED ISSUE #192: DaysUntilBirthday sample app requires keychain entitlement with Development certificate
Per [this conversation](https://github.com/google/GoogleSignIn-iOS/pull/191#discussion_r939049720) on #191 adding async/await support to the Swift sample app, we discovered that the `keychain-access-groups` entitlement is needed to sign in successfully. We should explore further why this is the case and if we can get around the issue.

mdmathias:
Fixes #192

Related to discussion in #191.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
+++ b/Samples/Swift/DaysUntilBirthday/macOS/DaysUntilBirthdayOnMac.entitlements
@@ -8,5 +8,7 @@
 	<true/>
 	<key>com.apple.security.network.client</key>
 	<true/>
+	<key>keychain-access-groups</key>
+	<array/>
 </dict>
 </plist>


petea:
Any insight into _why_ this is the behavior for the development cert case on macOS?

You didn't need to add keychain sharing with our development cert, correct?

mdmathias:
The issue is that running the sample app on macOS is, I guess, similar to deploying to a real device. That means there is a discrepancy between what the iOS and macOS sample apps need in order to work. iOS will not need a provisioning profile if deploying on the simulator. macOS will because you are deploying to a real device. I think this difference is a little confusing.

In any case, macOS users have two options:

1. Manually manage signing and provide a provisioning profile.
2. Add the `keychain-access-groups` entitlement to the mac target.

I think we should suggest in that order. I'm going to update our README to provide some guidance on this.

dav3yblaz3 (commented):
this looks like this could work thank you

PR #195, opened 2022-08-10
Fix Objc Sample App Build failure

Alex-4-Git:
Use the `completion` to replace `callback` according to the latest API change.


mdmathias (approved):

petea:
Good catch.

PR #196, opened 2022-08-10
 Move two methods from GIDAuthentication API to GIDGoogleUser API

Alex-4-Git:
1. Add two methods in GIDGoogleUser public API.
- @property(nonatomic, readonly)(id<GTMFetcherAuthorizationProtocol>) fetcherAuthorizer;
- (void)doWithFreshTokens:(void (^)(GIDGoogleUser *_Nullable user,
                                               NSError *_Nullable error))completion;

2. Remove GIDAuthentication.h from GSI public API.
3. Keep GIDAuthentication as an internal class.


GoogleSignIn/Sources/GIDAuthentication.h
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDAuthentication.h"
```
petea, line 17: At what point do you intend to complete the merge of `GIDAuthentication` into `GIDGoogleUser`?  Currently, both objects share the same `authState` reference, which creates complexity.

.gitignore
```diff
@@ -5,6 +5,7 @@
 *.pbxuser
 *.mode1v3
 *.mode2v3
+*.pbxproj
```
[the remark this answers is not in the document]
Alex-4-Git: Removed.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
```diff
@@ -53,6 +57,18 @@ NS_ASSUME_NONNULL_BEGIN
 /// see https://developers.google.com/identity/sign-in/ios/backend-auth.
 @property(nonatomic, readonly, nullable) GIDToken *idToken;
 
+/// Gets a new authorizer for `GTLService`, `GTMSessionFetcher`, or `GTMHTTPFetcher`.
```
petea, line 60: Update the doc comment to reflect that this will always return the same instance.

GoogleSignIn/Sources/GIDGoogleUser.m
```diff
@@ -103,36 +103,53 @@ - (GIDConfiguration *)configuration {
...
 - (nullable GIDToken *)idToken {
   @synchronized(self) {
-    NSString *idTokenString = _authState.lastTokenResponse.idToken;
-    if (!_cachedIdToken && idTokenString) {
+    return _cachedIdToken;
+  }
+}
+
+- (id<GTMFetcherAuthorizationProtocol>) fetcherAuthorizer {
+  return [_authentication fetcherAuthorizer];
```
petea, line 123: Let's look at creating a single `GTMAppAuthFetcherAuthorization` instance at `GIDGoogleUser` initialization time and letting it own the `OIDAuthState` data.  That instance could be stored in the property ivar and this getter would go away.

GoogleSignIn/Sources/GIDAuthentication.h
petea: I'm not sure how the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) is an argument for keeping `GIDAuthentication` here.  However, the general goal of [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) feels like a relevant consideration to me.  I don't think we want two classes that both read and mutate the same `OIDAuthState` object if we can avoid it.

Note that dropping `GIDAuthentication` and merging it into `GIDGoogleUser` was an explicit goal of our reviewed design for this refactoring of `GIDGoogleUser`.

GoogleSignIn/Sources/GIDAuthentication.m
```diff
@@ -255,15 +234,7 @@ - (void)doWithFreshTokens:(GIDAuthenticationCompletion)completion {
                                       callback:^(OIDTokenResponse *_Nullable tokenResponse,
                                                  NSError *_Nullable error) {
     if (tokenResponse) {
-      [self willChangeValueForKey:NSStringFromSelector(@selector(accessToken))];
```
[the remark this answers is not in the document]
petea: We should maintain the same level of testing coverage here.  Is there some reason this is particularly difficult?

GoogleSignIn/Sources/GIDAuthentication.h
Alex-4-Git: In our design our goal is to merge the public GIDAuthentication API into GIDGoogleUser API. We will accomplish that goal in this PR. And we keep GIDAuthentication class as private. I see most of your comments are about why not remove it. Let me try to explain it here.

Now the private GIDAuthentication class handles all authorization related methods. That is what I mean to follow single responsibility principle. Encapsulation, as you mentioned, is the main reason. We hide all the implementation details about handling EMM and talking to OIDAuthorizationService to refresh OIDAuthState. This will help us use fake to replace OCMock in the next project. In the next project we will rename GIDAuthentication to GIDAuthorizationService class and provide a fake. Please see doc Refactor GIDGoogleUser to improve tests.
https://docs.google.com/document/d/12lFJBakatD9KdKrBph3B0mw8Zw2KnVrkkwil6fFhB14/edit#

GoogleSignIn/Sources/GIDAuthentication.m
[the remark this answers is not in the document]
Alex-4-Git: This is not difficult. But these properties will change when OIDAuthentication `doWIthFreshTokens` method is invoked. That means we need to add new OCMocks. In the next project we will extract a new protocol and use fake s in unit tests. Then we will remove these mocks just added. I would prefer to avoid these back and forth changes.

GoogleSignIn/Sources/GIDGoogleUser.m
Alex-4-Git: As I mentioned in previous comment. GIDAuthentication is good encapsulation for authorization related methods.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
Alex-4-Git: This will change when the `authState` is updated by either refreshing tokens or adding new scopes.

petea:
Discussed offline and decided to resume this PR once we've documented the implementation approach for the larger GIDGoogleUser refactoring.

PR #198, opened 2022-08-16
Update CHANGELOG for 6.2.3 patch release

mdmathias:


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 6.2.3 (2022-8-18)
+- Fix resource loading in GoogleSignInSwift with CocoaPods use_frameworks! ([#197](https://github.com/google/GoogleSignIn-iOS/pull/197))
+- Exclude GoogleSignInSwift for i386 in addition to armv7. ([#166](https://github.com/google/GoogleSignIn-iOS/pull/166))
+
+# 6.2.2 (2022-5-27)
```
petea, line 5: I think you didn't mean to duplicate this section.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 6.2.3 (2022-8-18)
+- Fix resource loading in GoogleSignInSwift with CocoaPods use_frameworks! ([#197](https://github.com/google/GoogleSignIn-iOS/pull/197))
+- Exclude GoogleSignInSwift for i386 in addition to armv7. ([#166](https://github.com/google/GoogleSignIn-iOS/pull/166))
```
petea, line 3: "Prevent build errors for GoogleSignInSwift in certain scenarios when using Swift Package Manager."

I'd try to use a simplified version of the PR description here and then readers can use the link for more detail if desired.

petea (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 6.2.3 (2022-8-18)
```
petea, line 1: Wait to merge until we run the release process.

CHANGELOG.md
mdmathias: Ugh. Sorry about that.

CHANGELOG.md
mdmathias: 👍

CHANGELOG.md
mdmathias: I'll make the change. IMHO, I think it's a little confusing that the changelog bullet items don't match the linked commit. It would make me wonder if these link took me to the right place. This issue might suggest that some PR titles are too difficult to understand at a glance, in which case I'd argue that we should strive to make our titles understandable at a glance and fill out more detail in the descriptions.

What do you think?

petea: The visibility and audience are different between the changelog and PRs, so I think it can be helpful for them to diverge in some cases.

mdmathias: Agreed.

Issue #203, opened 2022-08-21
Fix document comment of GoogleSignIn (product "GoogleSignIn")


kamimi01:
## Description

I've realized that there is a difference between the display text of ".standard" and ".wide" when we use "GIDSignInButtonStyle" property.
But it took there is no document comments in `GIDSignInButton.h` about it.
I think that should be written in the source code.

w-goog:
Thanks for your contribution, @kamimi01! Regarding text customization - that's not a change the iOS library is planning to support at this time. If there's a gap in our localization, please let us know by reporting an issue here :)

Or, if you find one of the default buttons otherwise supported in the [brand guidelines](https://developers.google.com/identity/branding-guidelines) has specific issues on iOS, similarly, please open an issue.

PR #204, opened 2022-08-21
fix: Document comment for GIDSignInButtonStyle

kamimi01:
I worked on https://github.com/google/GoogleSignIn-iOS/issues/203

I think to be customizable the display text is better than adding document comments like this, so this is a workaround.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
@@ -25,9 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 ///
 /// The minimum size of the button depends on the language used for text.
 /// The following dimensions (in points) fit for all languages:
-/// - kGIDSignInButtonStyleStandard: 230 x 48
-/// - kGIDSignInButtonStyleWide:     312 x 48
-/// - kGIDSignInButtonStyleIconOnly: 48 x 48 (no text, fixed size)
+/// - kGIDSignInButtonStyleStandard: 230 x 48 (Display text is "Sign in")
+/// - kGIDSignInButtonStyleWide:     312 x 48 (Display text is "Sign in with Google")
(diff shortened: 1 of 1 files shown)


kamimi01:
@petea Hi team,
Thank you for checking my pull request. As I mentioned, I think to be customizable display text is better than adding doc comments. Do you have a plan to be able to customize it? If you don't, would you mind telling me a reason?
Thank you.

monicakh (approved):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
```diff
@@ -25,9 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 ///
 /// The minimum size of the button depends on the language used for text.
 /// The following dimensions (in points) fit for all languages:
-/// - kGIDSignInButtonStyleStandard: 230 x 48
-/// - kGIDSignInButtonStyleWide:     312 x 48
-/// - kGIDSignInButtonStyleIconOnly: 48 x 48 (no text, fixed size)
+/// - kGIDSignInButtonStyleStandard: 230 x 48 (Display text is "Sign in")
```
w-goog: Future-proofing this a bit :)

```suggestion
/// - kGIDSignInButtonStyleStandard: 230 x 48 (Example display text: "Sign in")
```

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignInButton.h
```diff
@@ -25,9 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// - kGIDSignInButtonStyleWide:     312 x 48 (Display text is "Sign in with Google")
```
w-goog:
```suggestion
/// - kGIDSignInButtonStyleWide:     312 x 48 (Example display text: "Sign in with Google")
```

PR #212, opened 2022-09-07
Add unit test for `updateAuthState:` in GIDGoogleUserTest

Alex-4-Git:
Our plan is to merge GIDAuthentication into GIDGoogleUser. This PR is to add more coverage to exiting GIDGoogleUser API and prepare for more changes in the coming PRs.

### DIFF
--- a/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
@@ -105,7 +105,7 @@ static NSTimeInterval const kTimeAccuracy = 10;
 }
 #endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
 
-- (void) testUpdateAuthState {
+- (void)testUpdateAuthState {
   NSString *idToken = [self idTokenWithExpireTime:kIDTokenExpireTime];
   OIDAuthState *authState = [OIDAuthState testInstanceWithIDToken:idToken
                                                       accessToken:kAccessToken
--- a/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
@@ -36,22 +36,12 @@
 #endif
 
 static NSString *const kNewAccessToken = @"new_access_token";
-static NSTimeInterval const kAccessTokenExpireTime = 442886100;
-static NSTimeInterval const kIDTokenExpireTime = 442886118;
-static NSTimeInterval const kNewAccessTokenExpireTime = 442886200;
-static NSTimeInterval const kNewIDTokenExpireTime = 442886224;
 static NSTimeInterval const kTimeAccuracy = 10;
 
 @interface GIDGoogleUserTest : XCTestCase
 @end
 
-@implementation GIDGoogleUserTest {
-  // Fake data used to generate the expiration date of the access token.
-  NSTimeInterval _accessTokenExpireTime;
-  
-  // Fake data used to generate the expiration date of the ID token.
-  NSTimeInterval _idTokenExpireTime;
-}
+@implementation GIDGoogleUserTest
 
 #pragma mark - Tests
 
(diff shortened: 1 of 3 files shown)


GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +110,64 @@ - (void)testLegacyCoding {
 }
 #endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
 
+- (void) testUpdateAuthState {
+  GIDGoogleUser *user = [self googleUser];
+  XCTAssertEqualObjects(user.accessToken.tokenString, kAccessToken);
+  [self assertDate:user.accessToken.expirationDate equalTime:_accessTokenExpireTime];
```
mdmathias, line 116: I think you're trying to make your tests DRY here, but sometimes that can add some mental overhead to reading and debugging failing tests. Consider removing this and [making your tests DAMP](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html).

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +110,64 @@ - (void)testLegacyCoding {
...
+      [OIDTokenResponse testInstanceWithIDToken:idToken
+                                    accessToken:kAccessToken
+                                      expiresIn:accessTokenExpiresIn
+                                   tokenRequest:tokenRequest];
+  return [[GIDGoogleUser alloc]
+          initWithAuthState:[OIDAuthState testInstanceWithTokenResponse:tokenResponse]
+                profileData:nil];
+}
+
+- (NSString *)idToken {
```
mdmathias, line 147: This and the getter below basically do the same thing with a different expiration time, right? Do you need these methods? If you're using these methods as a kind of named storage, consider using a property (or ivar) for each to make things clearer.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +105,40 @@ - (void)testLegacyCoding {
...
+  XCTAssertEqualObjects(user.idToken.tokenString, idToken);
+  XCTAssertEqualWithAccuracy([user.idToken.expirationDate timeIntervalSinceReferenceDate],
+                             kIDTokenExpireTime, kTimeAccuracy);
+  
+  NSString *idTokenNew = [self idTokenWithExpireTime:kNewIDTokenExpireTime];
+  OIDAuthState *newAuthState = [OIDAuthState testInstanceWithIDToken:idTokenNew
+                                                         accessToken:kNewAccessToken
+                                               accessTokenExpireTime:kNewAccessTokenExpireTime];
+  
+  [user updateAuthState:newAuthState profileData:nil];
```
petea, line 128: How about testing the case where `profileData` is provided?

Alex-4-Git: Done.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -34,10 +35,23 @@
 #import <AppAuth/OIDAuthState.h>
 #endif
 
+static NSString *const kNewAccessToken = @"new_access_token";
+static NSTimeInterval const kAccessTokenExpireTime = 442886100;
+static NSTimeInterval const kIDTokenExpireTime = 442886118;
+static NSTimeInterval const kNewAccessTokenExpireTime = 442886200;
```
mdmathias, line 41: I don't think these constants are used outside of `testUpdateAuthState` below. Do you have plans to use them in future tests? If not, then maybe they don't need to be constants right now? 

This comment is driven by my concern that the constants don't really have meaningful names (or meaning in general) outside of `testUpdateAuthState`. 

You could create `NSTimeInterval`s within the method and give them good names in the new test method. Then, you could create new intervals with better names than "new" (e.g., something like `updatedAccessTokenExpireTime`) in the test method. 

[NSDate has a bunch of methods](https://developer.apple.com/documentation/foundation/date/1779636-timeintervalsince) that you can use to create time intervals.

PS I get the impression that you're following a pre-existing pattern in [GIDAuthenticationTest](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Tests/Unit/GIDAuthenticationTest.m), but IMHO it's not very readable there either.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -34,10 +35,23 @@
...
+static NSTimeInterval const kNewIDTokenExpireTime = 442886224;
+static NSTimeInterval const kTimeAccuracy = 10;
+
 @interface GIDGoogleUserTest : XCTestCase
 @end
 
-@implementation GIDGoogleUserTest
+@implementation GIDGoogleUserTest {
+  // Fake data used to generate the expiration date of the access token.
+  NSTimeInterval _accessTokenExpireTime;
```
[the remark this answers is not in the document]
Alex-4-Git: Removed.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
Alex-4-Git: These value will be used in other test cases. But I think creating them inside the tests make the code easier to understand. Thanks for pointing it out.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +110,64 @@ - (void)testLegacyCoding {
 }
 #endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
 
+- (void) testUpdateAuthState {
+  GIDGoogleUser *user = [self googleUser];
```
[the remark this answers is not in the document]
mdmathias: I see. Resolving.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +95,48 @@ - (void)testLegacyCoding {
 }
 #endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
 
+- (void)testUpdateAuthState {
+  NSTimeInterval accessTokenExpireTime = [NSDate timeIntervalSinceReferenceDate];
+  NSTimeInterval idTokenExpireTime = [NSDate timeIntervalSinceReferenceDate] + 100;
```
mdmathias, line 100: Do 100, 200, and 300 have special meaning, or are you just making sure that these times are different? I _think_ the time intervals should be different with each subsequent call to `[NSDate timeIntervalSinceReferenceDate]`. If so, maybe just drop the addition.

mdmathias (approved):

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
petea: @Alex-4-Git could you respond to Matt's question here.

I'd also suggest marking comment threads as resolved when you have addressed them so as not to miss those that still need to be addressed.

Alex-4-Git: In the test we use a method `XCTAssertEqualWithAccuracy`. We want to make sure the difference between the expire times is larger than the accuracy so they would not be misused. 

I have used a constant kTimeIncrement to improve the readability.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -91,4 +98,40 @@ - (void)testLegacyCoding {
...
+  XCTAssertEqualObjects(user.idToken.tokenString, updatedIDToken);
+  XCTAssertEqualWithAccuracy([user.idToken.expirationDate timeIntervalSinceReferenceDate],
+                             updatedIDTokenExpireTime, kTimeAccuracy);
+  XCTAssertEqual(user.profile, updatedProfileData);
+}
+
+#pragma mark - Helpers
+
+- (NSString *)idTokenWithExpireTime:(NSTimeInterval)expireTime {
+  return [OIDTokenResponse idTokenWithSub:kUserID exp:@(expireTime + NSTimeIntervalSince1970)];
```
mdmathias, line 134: Why does this method use `expireTime + NSTimeIntervalSince1970`? Removing `NSTimeIntervalSince1970` causes the tests to fail, which is confusing to me. 

Also, this method name suggests that it makes an ID token with the given `expireTime`, but what it does is create an ID token with the given expire time plus this extra increment. Maybe update the name and add a comment to clarify what it is doing and why.

Alex-4-Git: This is the method we used to generate an id token in GIDAuthenticationTest. We can figure it out how it works offline and improve it in a separate PR.

petea: > Why does this method use `expireTime + NSTimeIntervalSince1970`? Removing `NSTimeIntervalSince1970` causes the tests to fail, which is confusing to me.

The `NSTimeInterval` values generated by [`timeIntervalSinceReferenceDate`](https://developer.apple.com/documentation/foundation/nsdate/1408189-timeintervalsincereferencedate?language=objc) are relative to the reference date of Jan 1, 2001.  Here, [`NSTimeIntervalSince1970`](https://developer.apple.com/documentation/foundation/nstimeintervalsince1970?language=objc) is being used to convert it to a unix epoch value, which is the reference date used by the tokens.  We could probably simplify by using [`timeIntervalSince1970`](https://developer.apple.com/documentation/foundation/nsdate/1407504-timeintervalsince1970?language=objc) instead.

mdmathias: Thanks for clarifying. I had a feeling that was the case. I agree that simplifying would be better

PR #219, opened 2022-09-13
Ignore warnings from GTMFetcherAuthorizationProtocol deprecation

petea:
With the release of [GTMAppAuth 1.3.1](https://github.com/google/GTMAppAuth/releases/tag/1.3.1), we're now building against GTMSessionFetcher 2.1.x by default and need to handle warnings stemming from the [deprecation of `GTMFetcherAuthorizationProtocol`](https://github.com/google/gtm-session-fetcher/pull/315).

### DIFF
--- a/GoogleSignIn/Sources/GIDAuthentication.m
+++ b/GoogleSignIn/Sources/GIDAuthentication.m
@@ -133,9 +133,12 @@ static NSString *const kNewIOSSystemName = @"iOS";
   GTMAppAuthFetcherAuthorizationEMMChainedDelegate *chainedDelegate =
       [[GTMAppAuthFetcherAuthorizationEMMChainedDelegate alloc] initWithDelegate:delegate
                                                                         selector:sel];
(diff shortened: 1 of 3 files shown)


mdmathias (commented):
Should we update `GTMAppAuth` to conform to `GTMSessionFetcherAuthorizer` so that we don't need to worry about the deprecation of `GTMFetcherAuthorizationProtocol`?

I guess that change would require a major version release of `GTMAppAuth`?

petea:
> Should we update `GTMAppAuth` to conform to `GTMSessionFetcherAuthorizer` so that we don't need to worry about the deprecation of `GTMFetcherAuthorizationProtocol`?
> 
> I guess that change would require a major version release of `GTMAppAuth`?

We should consider this for our next major release.  We'll have to decide if we're comfortable dropping support for anything before GTMSessionFetcher 2.1.0.

mdmathias (approved):

PR #221, opened 2022-09-15
Create separate integration test workflow

mdmathias:


petea:
Should we removed the disabled job from tests.yml?

mdmathias:
@petea Yes, sorry for that. Removed now.

petea (approved):

PR #229, opened 2022-09-26
Use accessTokenExpiresIn to replace accessTokenExpireTime in tests

Alex-4-Git:
OIDTokenResponse requires a time interval to indicate the life time of the access token such as 3600.

In the current unit tests we first set the expiration time using 
`NSTimeInterval accessTokenExpireTime = [[NSDate date] timeIntervalSince1970] + accessExpireIn;`
Then when we need the time interval we do 
`NSNumber *accessTokenExpiresIn = @(accessTokenExpireTime - [[NSDate date] timeIntervalSince1970]);`

This PR is to simplify this by directly set the life time of accessToken `accessTokenExpiresIn`.


GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -155,12 +154,12 @@ - (void)testUpdateAuthStateWithUnchangedRefreshTokenAndIDToken {
 
 #pragma mark - Helpers
 
-- (GIDGoogleUser *)googleUserWithAccessTokenExpireTime:(NSTimeInterval)accessTokenExpireTime
-                                     idTokenExpireTime:(NSTimeInterval)idTokenExpireTime {
+- (GIDGoogleUser *)googleUserWithAccessTokenExpiresIn:(NSTimeInterval)accessTokenExpiresIn
+                                    idTokenExpireTime:(NSTimeInterval)idTokenExpireTime {
```
petea, line 158: Should we make ID token expiration consistent with access token here?

Alex-4-Git: I prefer to make it simple. The less back and forth conversion, the less likely there are mistakes.

petea: I don't feel strongly about it, as it's just a test helper, but I think it might be easier to work with if the signature were:

```
- (GIDGoogleUser *)googleUserWithAccessTokenExpiresIn:(NSTimeInterval)accessTokenExpiresIn
                                     idTokenExpiresIn:(NSTimeInterval)idTokenExpiresIn
```

petea (approved):

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
Alex-4-Git: That's a brilliant idea.

PR #237, opened 2022-10-11
Future-proof against some upcoming fetcher changes.

mwyman:
To eliminate `dispatch_semaphore` usage, the `GTMSessionFetcherService` protocol was changed recently; this adds the required method to ensure the tests don't break if building with the latest fetcher.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -5,14 +5,14 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   workflow_dispatch:
 
 jobs:
 
   swift-button-functional-test:
-    runs-on: macOS-latest
+    runs-on: macOS-12
+    # Don't run if triggered by a PR from a fork since our Secrets won't be provided to the runner.
+    if: "!github.event.pull_request.head.repo.fork"
     defaults:
       run:
         working-directory: Samples/Swift/DaysUntilBirthday
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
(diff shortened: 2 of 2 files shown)


GoogleSignIn/Tests/Unit/GIDFakeFetcherService.m
```diff
@@ -63,6 +63,15 @@ - (NSURLSession *)sessionForFetcherCreation {
   return nil;
 }
 
+// TODO: replace with GTMSessionFetcherSessionCreationBlock once that has become
+// available by updating the version dependencies.
+typedef NSURLSession *_Nullable (^GIDFakeFetcherSessionCreationBlock)(
```
petea, line 68: Maybe move the typedef above for readability?

GoogleSignIn/Tests/Unit/GIDFakeFetcherService.m
```diff
@@ -63,6 +63,15 @@ - (NSURLSession *)sessionForFetcherCreation {
...
+    id<NSURLSessionDelegate> _Nullable delegate);
+- (NSURLSession *)sessionWithCreationBlock:
```
petea, line 70: Might add comments to the effect of "this is for pre-release support" since we don't typically build against your HEAD.

GoogleSignIn/Tests/Unit/GIDFakeFetcherService.m
mwyman: The upcoming fetcher changes have been tweaked, the end result being that rather than adding a new method, the fetcher is moving most of the methods out of the protocol; but then the fake fetcher service has some `@synthesize` properties that aren't defined (until the fetcher update gets pushed as a release).

PR #238, opened 2022-10-14
Fix the signInSample app

Alex-4-Git:
Since we refactored the GIDGoogleUser API we need to update the property name list in the `AuthInspectorViewController`. Otherwise there will be no content displayed in the view.


mdmathias:
@Alex-4-Git could you add some context on what this PR fixes in the description?

mdmathias (approved):

PR #246, opened 2022-10-31
Improve documentation for method `handleURL:` in GIDSignIn public API

Alex-4-Git:


GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -80,6 +80,8 @@ typedef void (^GIDDisconnectCompletion)(NSError *_Nullable error);
 /// :nodoc:
 - (instancetype)init NS_UNAVAILABLE;
 
+/// Handles the URL returned from SFSafariViewController for iOS 9 and 10.
```
petea, line 83: This is accurate but doesn't tell the full story.  The custom URL scheme is used for app flipping in the MDM flow on all iOS versions.  We don't want developers to think they don't need to handle the custom URL scheme if they don't support iOS 9 and 10.

Alex-4-Git: Thanks, I have updated the comments.
What is app flipping?

Alex-4-Git: The app flipping means the communications between GSI and Device policy app. Clients should always handle custom URL and don't need to know the details. 

We should move this comments into our implementations for records.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -169,6 +169,10 @@ @implementation GIDSignIn {
 
 #pragma mark - Public methods
 
+// Handles the custom URL returned from AppAuth or Device Policy App (EMM flow).
```
mdmathias, line 172: I think you need three slashes: `///`.

mdmathias, line 172: I think also it would be more clear to use a different word than 'handle' in the description since that is the word used in the method.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
Alex-4-Git: I tried this. I think this only works for comments in the header file. The comments in the source file would not be displayed.

Alex-4-Git: Replaced with `open`.

mdmathias: Gah, of course. What do you think about moving this to `GIDSignIn.h` and merging this information with the existing doc comment?

Alex-4-Git: This is the implementation details for us. Our clients should not know this.

mdmathias: Since this is a public method, don't these comments apply more broadly than "implementation details" implies?

Alex-4-Git: We have the comments in the public API for the clients. They don't need to understand how we use this method in our implementation.

mdmathias: Perhaps I'm not understanding things correctly, so please bear with me. :) 

My thinking is that this method is in the public API. For example, our [sample app](https://github.com/google/GoogleSignIn-iOS/blob/8345fcc43c38cefb67f1ad0d5dcd51a1d036e460/Samples/ObjC/SignInSample/Source/AppDelegate.m#L49) uses this method much the same way clients will. 

Therefore, it seems to me that these comments/details could relevant to others.

mdmathias: Following up from an offline conversation...

It seems like there was already a discussion about this that led to these comments landing here. Is that right @petea? If so, I don't want to block this PR on a small detail. I'll approve and @Alex-4-Git you can decide what to do!

mdmathias (approved):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -170,6 +170,10 @@ @implementation GIDSignIn {
 
 #pragma mark - Public methods
 
+// Opens the custom URL returned from AppAuth or Device Policy App (EMM flow).
```
petea, line 173:
```
// Handles the custom scheme URL opened by SFSafariViewController or the Device Policy App.
//
// For SFSafariViewController invoked via AppAuth, this method is used on iOS 10.
// For the Device Policy App (EMM flow) this method is used on all iOS versions.
```

petea (approved):

GoogleSignIn/Sources/GIDSignIn.m
Alex-4-Git: Peter helped me improve the comments. We can put it in the source file.

PR #247, opened 2022-11-01
Improve our workflows

petea:
This addresses three issues:

- The `pull_request` trigger is currently scoped to `main` branch only, so jobs aren't run by default for PRs on feature branches.
- Warnings about the`macOS-latest` runner alias transitioning from `macOS-11` to `macOS-12`.
- Integration tests failing when triggered by a PR from a fork due to the secret not being passed to the runner in [this scenario](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories).

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -11,7 +11,7 @@ jobs:
 
   swift-button-functional-test:
     runs-on: macOS-12
-    if: !github.event.pull_request.head.repo.fork
+    if: "!github.event.pull_request.head.repo.fork"
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


s8627:
- [ ] ?

Alex-4-Git (approved):

Alex-4-Git:
Provide some descriptions.

petea:
> Provide some descriptions.

Added a PR description.

.github/workflows/integration_tests.yml
```diff
@@ -5,14 +5,13 @@ on:
...
-    branches:
-      - main
   workflow_dispatch:
 
 jobs:
 
   swift-button-functional-test:
-    runs-on: macOS-latest
+    runs-on: macOS-12
+    if: "!github.event.pull_request.head.repo.fork"
```
mdmathias, line 14: Is it worth adding a comment about why this is needed? The code isn't necessarily self-documenting and future self may not know or remember what issue is being addressed here.

mdmathias (approved):
Just a note to leave a clarifying comment. LGTM otherwise.

.github/workflows/integration_tests.yml
petea: For sure.  Done.

PR #261, opened 2022-11-29
Add the structure for the protocol GIDKeychainHandlling and implementations

Alex-4-Git:
We can use this PR to decide how to structure protocols and implementations.

I suggest that for each protocol we can create a folder whose name is the protocol like `GIDKeychainHandlling`. In the folder we can create the protocol and implementations.


GoogleSignIn/Sources/GIDKeychainHandlling/GIDKeychainHandlling.h
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+@class OIDAuthState;
+
+@protocol GIDKeychainHandlling <NSObject>
```
petea, line 21: `GIDKeychainHandlling` -> `GIDKeychainHandling`

petea (changes requested):

PR #259, opened 2022-11-24
Fix macro in GIDGoogleUser.m

Alex-4-Git:
Fix the the build integration test failure on GSI-iOS.
https://github.com/google/GoogleSignIn-iOS/actions/runs/3502685894/jobs/5867240619#step:4:1011

PR #263, opened 2022-11-29
Follow up: Fix macro in GIDGoogleUser.m 

Alex-4-Git:
Follow up with PR 259 to fix the build integration test failure on GSI-iOS.
https://github.com/google/GoogleSignIn-iOS/actions/runs/3502685894/jobs/5867240619#step:4:1011

https://github.com/google/GoogleSignIn-iOS/pull/259


petea:
@Alex-4-Git have you confirmed the fix by building one of the sample apps for Mac Catalyst?

Alex-4-Git:
Reproduced the failure and tested the fix on mac catalyst.

petea (approved):

PR #267, opened 2022-12-07
Remove `GIDDisconnectCompletion`

petea:
For consistency with the rest of our completion block typedef removals from public headers.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
@@ -51,9 +51,6 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
   kGIDSignInErrorCodeMismatchWithCurrentUser = -9,
 };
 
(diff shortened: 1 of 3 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -51,9 +51,6 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
   kGIDSignInErrorCodeMismatchWithCurrentUser = -9,
 };
 
-/// Represents a completion block that takes an error if the operation was unsuccessful.
-typedef void (^GIDDisconnectCompletion)(NSError *_Nullable error);
-
```
Alex-4-Git, line 56: Maybe we can keep this in `GIDSignIn_Private.h`?

petea: Sure, this would be consistent with what we've done for `GIDUserAuthCompletion`.  Added to private header.

mdmathias (approved):

PR #268, opened 2022-12-08
Rename GIDUserAuth to GIDSignInResult

petea:


Alex-4-Git:
Why do you want to change the name of `UserAuth`? I think we use this name because it is the result of sign-in and adding scopes.

petea:
> Why do you want to change the name of `UserAuth`? I think we use this name because it is the result of sign-in and adding scopes.

As I mentioned in the PR description, I don't feel like the name `GIDUserAuth` does a good job conveying the roll of the class.

mdmathias (approved):

PR #280, opened 2023-01-04
Move authorizationEndpointURL and tokenEndpointURL into GIDSignInPreferences.

Alex-4-Git:
This is the prep PR for GIDAuthorizationFlowProcessor. In order to start the authorization flow GIDAuthorizationFlowProcessor makes an OIDAuthorizationRequest to AppAuth which requires the appAuthConfiguration object.

This PR decouples appAuthConfiguration with GIDSignIn so it could be moved into GIDAuthorizationFlowProcessor implementation.

### DIFF
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -462,12 +462,22 @@ static NSString *const kConfigOpenIDRealmKey = @"GIDOpenIDRealm";
 - (id)initPrivate {
   id<GIDKeychainHandler> keychainHandler = [[GIDKeychainHandler alloc] init];
   id<GIDHTTPFetcher> httpFetcher = [[GIDHTTPFetcher alloc] init];
+  
+  NSString *authorizationEnpointURL = [NSString stringWithFormat:kAuthorizationURLTemplate,
+      [GIDSignInPreferences googleAuthorizationServer]];
+  NSString *tokenEndpointURL = [NSString stringWithFormat:kTokenURLTemplate,
+      [GIDSignInPreferences googleTokenServer]];
+  OIDServiceConfiguration *appAuthConfiguration = [[OIDServiceConfiguration alloc]
+      initWithAuthorizationEndpoint:[NSURL URLWithString:authorizationEnpointURL]
+                      tokenEndpoint:[NSURL URLWithString:tokenEndpointURL]];
   return [self initWithKeychainHandler:keychainHandler
-                           httpFetcher:httpFetcher];
+                            httpFetcher:httpFetcher
+                   appAuthConfiguration:appAuthConfiguration];
 }
 
 - (instancetype)initWithKeychainHandler:(id<GIDKeychainHandler>)keychainHandler
-                            httpFetcher:(id<GIDHTTPFetcher>)httpFetcher{
+                            httpFetcher:(id<GIDHTTPFetcher>)httpFetcher
+                   appAuthConfiguration:(nullable OIDServiceConfiguration *)appAuthConfiguration {
   self = [super init];
   if (self) {
     // Get the bundle of the current executable.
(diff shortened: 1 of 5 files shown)


GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -462,12 +462,22 @@ + (GIDSignIn *)sharedInstance {
...
   id<GIDKeychainHandler> keychainHandler = [[GIDKeychainHandler alloc] init];
   id<GIDHTTPFetcher> httpFetcher = [[GIDHTTPFetcher alloc] init];
+  
+  NSString *authorizationEnpointURL = [NSString stringWithFormat:kAuthorizationURLTemplate,
+      [GIDSignInPreferences googleAuthorizationServer]];
+  NSString *tokenEndpointURL = [NSString stringWithFormat:kTokenURLTemplate,
+      [GIDSignInPreferences googleTokenServer]];
+  OIDServiceConfiguration *appAuthConfiguration = [[OIDServiceConfiguration alloc]
+      initWithAuthorizationEndpoint:[NSURL URLWithString:authorizationEnpointURL]
+                      tokenEndpoint:[NSURL URLWithString:tokenEndpointURL]];
```
ch40w31, line 472: The initialization seems pretty static, so it can be placed anywhere essentially. And it feels like it should be just an implementation detail of the proposed `GIDAuthorizationFlowProcessor` unless it will still be needed outside `GIDAuthorizationFlowProcessor`.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -462,12 +462,22 @@ + (GIDSignIn *)sharedInstance {
...
   return [self initWithKeychainHandler:keychainHandler
-                           httpFetcher:httpFetcher];
+                            httpFetcher:httpFetcher
+                   appAuthConfiguration:appAuthConfiguration];
```
ch40w31, line 475: Check colon alignment.

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
Alex-4-Git: It will also be used in GIDSignIn.m here.

[GIDAuthStateMigration migrateIfNeededWithTokenURL:_appAuthConfiguration.tokenEndpoint
                                          callbackPath:kBrowserCallbackPath
                                          keychainName:kGTMAppAuthKeychainName
                                        isFreshInstall:isFreshInstall];

We can also consider generating the token endpoint again to get rid of `_appAuthConfiguration` but I feel that makes our maintenance harder.

GoogleSignIn/Sources/GIDSignIn.m
Alex-4-Git: Done.

GoogleSignIn/Sources/GIDSignIn.m
ch40w31: It looks like `GIDAuthStateMigration` only needs the token URL. In that case, I'd suggest putting the token URL in a common utility class like `GIDSignInPreferences`, and move initialization of `OIDServiceConfiguration` into `GIDAuthorizationFlowProcessor`.

Alex-4-Git: Done.

GoogleSignIn/Sources/GIDSignInPreferences.m
```diff
@@ -94,6 +100,18 @@ + (NSString *)googleUserInfoServer {
   return kUserInfoServer;
 }
 
++ (NSURL *)authorizationEndpointURL {
+  NSString *authorizationEnpointURL = [NSString stringWithFormat:kAuthorizationURLTemplate,
+      [GIDSignInPreferences googleAuthorizationServer]];
```
ch40w31, line 105: You can do `[self googleAuthorizationServer]` instead, which is slightly preferable as it doesn't require an update if the class ever gets renamed.

GoogleSignIn/Sources/GIDSignInPreferences.m
```diff
@@ -94,6 +100,18 @@ + (NSString *)googleUserInfoServer {
...
+  return [NSURL URLWithString:authorizationEnpointURL];
+}
+
++ (NSURL *)tokenEndpointURL {
+  NSString *tokenEndpointURL = [NSString stringWithFormat:kTokenURLTemplate,
+      [GIDSignInPreferences googleTokenServer]];
```
ch40w31, line 111: Same here, `[self googleTokenServer]`.

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDSignInPreferences.m
Alex-4-Git: Done.

ch40w31 (approved):

PR #299, opened 2023-03-27
Use GTMAppAuth's new delegate protocol

mdmathias:
[GTMAppAuth added a new protocol](https://github.com/google/GTMAppAuth/pull/224) to provide an interface for dependencies to both add additional token refresh parameters and update errors issued by GTMAppAuth. The latter is especially for the case of EMM error handling.

Fixes https://github.com/google/GoogleSignIn-iOS/issues/304


GoogleSignIn/Tests/Unit/GIDFailingOIDAuthState.m
```diff
@@ -0,0 +1,52 @@
...
+#import "GoogleSignIn/Tests/Unit/OIDAuthState+Testing.h"
+#import "GoogleSignIn/Tests/Unit/OIDAuthorizationResponse+Testing.h"
+#import "GoogleSignIn/Tests/Unit/OIDTokenResponse+Testing.h"
+
+@implementation GIDFailingOIDAuthState
+
++ (instancetype)testInstance {
+  return [[GIDFailingOIDAuthState alloc]
+            initWithAuthorizationResponse:[OIDAuthorizationResponse testInstance]
+            tokenResponse:[OIDTokenResponse testInstance]];
```
ch40w31, line 27: Check indentation

GoogleSignIn/Tests/Unit/GIDFakeFetcher.m
```diff
@@ -39,6 +57,15 @@ - (void)beginFetchWithCompletionHandler:(FetchCompletionHandler)handler {
     [NSException raise:NSInvalidArgumentException format:@"Attempted start fetch again"];
   }
   _handler = [handler copy];
+  [self authorizeRequestWithCompletion:handler];
+}
+
+- (void)authorizeRequestWithCompletion:(FetchCompletionHandler)completion {
+  NSMutableURLRequest *mutableRequest = [NSMutableURLRequest requestWithURL:self.request.URL];
+  [self.authorizer authorizeRequest:mutableRequest completionHandler:^(NSError * _Nullable error) {
+    self.fetcherError = error;
```
ch40w31, line 66: Is `fetcherError` used after being assigned here?

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -47,7 +49,7 @@
 #import <AppAuth/OIDTokenResponse.h>
 #import <GoogleUtilities/GULSwizzler.h>
 #import <GoogleUtilities/GULSwizzler+Unswizzle.h>
-#import <GTMAppAuth/GTMAppAuthFetcherAuthorization.h>
+//#import <GTMAppAuth/GTMAppAuthFetcherAuthorization.h>
```
ch40w31, line 52: Remove

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -57,6 +63,40 @@ @interface GIDEMMSupportTest : XCTestCase
...
+  authSession.delegate = emmSupport;
+  GIDFakeFetcherService *fakeFetcherService = [[GIDFakeFetcherService alloc]
+                                                initWithAuthorizer:authSession];
+
+  NSURLRequest *request = [NSURLRequest requestWithURL:[NSURL URLWithString:@""]];
+  GTMSessionFetcher *fakeFetcher = [fakeFetcherService fetcherWithRequest:request];
+
+  GIDGoogleUser *user = [[GIDGoogleUser alloc] init];
+  GIDTestWorker *testWorker = [[GIDTestWorker alloc] initWithGoogleUser:user
+                                                                fetcher:fakeFetcher];
```
ch40w31, line 82: Not sure `GIDTestWorker` is necessary.

`GIDGoogleUser` is initialized using `init`, which isn't overridden in `GIDGoogleUser`, so it doesn't connect the instance with `GIDEMMSupport`. And `GIDTestWorker` does not do anything to the input `user` either.

Why not just call `[fakeFetcher beginFetchWithCompletionHandler:]` directly?

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDAuthStateMigration.h
```diff
@@ -16,18 +16,34 @@
...
 @interface GIDAuthStateMigration : NSObject
 
-// Perform a one-time migration for auth state saved by GPPSignIn 1.x or GIDSignIn 1.0 - 4.x to the
-// GTMAppAuth storage introduced in GIDSignIn 5.0.
-+ (void)migrateIfNeededWithTokenURL:(NSURL *)tokenURL
+/// Creates an instance of this migration type.
+///
+/// - Note: Uses `"auth"` as the keychain item name for the `GTMKeychainStore`, which is created
+/// within the initializer.
+- (instancetype)init;
```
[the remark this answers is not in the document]
mdmathias: I don't believe so at this point. Just added for convenience.

GoogleSignIn/Sources/GIDAuthStateMigration.h
```diff
@@ -16,18 +16,34 @@
...
+/// Perform a one-time migration for auth state saved by GPPSignIn 1.x or GIDSignIn 1.0 - 4.x to the
+/// GTMAppAuth storage introduced in GIDSignIn 5.0.
+- (void)migrateIfNeededWithTokenURL:(NSURL *)tokenURL
                        callbackPath:(NSString *)callbackPath
                        keychainName:(NSString *)keychainName
                      isFreshInstall:(BOOL)isFreshInstall;
 
+/// Returns a `GTMAuthSession` given the provided token URL.
+- (nullable GTMAuthSession *)
+    extractAuthorizationWithTokenURL:(NSURL *)tokenURL callbackPath:(NSString *)callbackPath;
```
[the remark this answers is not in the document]
mdmathias: It's used in some tests. It's still internal to the library and not part of the public API.

GoogleSignIn/Sources/GIDEMMSupport.h
```diff
@@ -20,19 +20,25 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
-// A class to support EMM (Enterprise Mobility Management).
-@interface GIDEMMSupport : NSObject
+/// A class to support EMM (Enterprise Mobility Management).
+@interface GIDEMMSupport : NSObject<GTMAuthSessionDelegate>
+
+- (instancetype)init NS_DESIGNATED_INITIALIZER;
+
++ (nullable NSError *)handleTokenFetchEMMError:(nullable NSError *)error;
```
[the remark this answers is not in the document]
mdmathias: Good catch.

Samples/Swift/DaysUntilBirthday/Podfile
```diff
@@ -2,6 +2,8 @@ pod 'GoogleSignIn', :path => '../../../', :testspecs => ['unit']
 pod 'GoogleSignInSwiftSupport', :path => '../../../', :testspecs => ['unit']
 project 'DaysUntilBirthdayForPod.xcodeproj'
 
+use_frameworks! :linkage => :static
```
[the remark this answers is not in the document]
mdmathias: It's related to GTMAppAuth being translated to Swift and is listed in the [3.0.0 release notes](https://github.com/google/GTMAppAuth/releases/tag/3.0.0).

More info in [CocoaPods 1.5.0 release announcement](https://blog.cocoapods.org/CocoaPods-1.5.0/) and in [the documentation](https://guides.cocoapods.org/syntax/podfile.html).

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
mdmathias: Sorry this test was unclear. I've made some changes to hopefully make things more clear and simultaneously closer to the intended use-case (by our clients, who will use `-[GIDGoogleUser fetcherAuthorizer]` to pass to their own `GTMSessionFetcher` instances.

The goal for `GIDTestWorker` was to simulate some toy class that a client may have that does some work. In this case, all it does is fail, but that is by design so that we can test (and document) the path by which an EMM error case occurs and is handled.

Please take a look and comment or resolve as appropriate.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -57,6 +63,35 @@ @interface GIDEMMSupportTest : XCTestCase
...
+
+  GIDFailingOIDAuthState *authState = [GIDFailingOIDAuthState testInstance];
+  GIDGoogleUser *user = [[GIDGoogleUser alloc] initWithAuthState:authState profileData:nil];
+  GIDFakeFetcherService *fakeFetcherService = [[GIDFakeFetcherService alloc]
+                                                initWithAuthorizer:user.fetcherAuthorizer];
+
+  NSURLRequest *request = [NSURLRequest requestWithURL:[NSURL URLWithString:@""]];
+  GTMSessionFetcher *fakeFetcher = [fakeFetcherService fetcherWithRequest:request];
+
+  GIDTestWorker *testWorker = [[GIDTestWorker alloc] initWithFetcher:fakeFetcher];
```
ch40w31, line 77: Thanks for updating this test. However, I still feel the `GIDTestWorker` isn't necessary -- it actually makes it a bit more difficult to read the code because the reader needs to jump between this file and GIDTestWorker.m to understand this test case.

One benefit of having `GIDTestWorker` is that it makes it clear that the "work" is going to fail by naming the method `failWorkWithCompletion:`. I think this is already implied by initializing `GIDGoogleUser` with a `GIDFailingOIDAuthState`. If you want to make this more explicit, you can add a comment right before calling `[fakeFetcher failWorkWithCompletion:]`.

ch40w31 (changes requested):

.github/workflows/unit_tests.yml
```diff
@@ -16,11 +16,6 @@ jobs:
       matrix:
         os: [macos-11, macos-12]
         podspec: [GoogleSignIn.podspec, GoogleSignInSwiftSupport.podspec]
-        flag: [
-          "", 
-          "--use-libraries", 
-          "--use-static-frameworks"
```
paulb777, line 22: `--use-static-frameworks` should still pass

GoogleSignIn.podspec
```diff
@@ -32,8 +33,8 @@ The Google Sign-In SDK allows users to sign in with their Google account from th
   ]
   s.ios.framework = 'UIKit'
   s.osx.framework = 'AppKit'
-  s.dependency 'AppAuth', '~> 1.5'
-  s.dependency 'GTMAppAuth', '>= 1.3', '< 3.0'
+  s.dependency 'AppAuth', '~> 1.6'
+  s.dependency 'GTMAppAuth', '~> 3.0.0'
```
paulb777, line 37: Should be 3.0 to allow minor (non-breaking) version updates.

`~> 3.0.0` only allows patch version updates.

Package.swift
```diff
@@ -44,11 +44,11 @@ let package = Package(
     .package(
       name: "AppAuth",
       url: "https://github.com/openid/AppAuth-iOS.git",
-      "1.5.0" ..< "2.0.0"),
+      "1.6.0" ..< "2.0.0"),
     .package(
       name: "GTMAppAuth",
       url: "https://github.com/google/GTMAppAuth.git",
-      "1.3.0" ..< "3.0.0"),
+      from: "3.0.0"),
```
paulb777, line 51: Should cap at < 4.0.0 so clients don't get breaking changes - and to be consistent with  CocoaPods

paulb777 (commented):
CocoaPods and SPM config changes LGTM.

GoogleSignIn/Tests/Unit/GIDFakeFetcher.m
mdmathias: Gah no it isn't and it is not needed.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
mdmathias: I've removed `GIDTestWorker`.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
mdmathias: Resolving since I've removed `GIDTestWorker` and have renamed some things in the test to make the intention more clear.

GoogleSignIn/Sources/GIDGoogleUser_Private.h
```diff
@@ -31,17 +31,16 @@ NS_ASSUME_NONNULL_BEGIN
 /// A completion block that takes a `GIDGoogleUser` or an error if the attempt to refresh tokens was unsuccessful.
 typedef void (^GIDGoogleUserCompletion)(GIDGoogleUser *_Nullable user, NSError *_Nullable error);
 
-// Internal methods for the class that are not part of the public API.
-@interface GIDGoogleUser () <GTMAppAuthFetcherAuthorizationTokenRefreshDelegate,
-                             OIDAuthStateChangeDelegate>
+/// Internal methods for the class that are not part of the public API.
+@interface GIDGoogleUser () <GTMAuthSessionDelegate, OIDAuthStateChangeDelegate>
```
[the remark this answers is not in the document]
mdmathias: Removed the declare protocol conformance. It was a vestige of past code.

The relationship is similar to [what was previously in place](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDGoogleUser.m#L232-L234). Previously, if on iOS and if EMM was supported, we would create an instance of a `GTMAppAuthFetcherAuthorization` [subclass within GoogleSignIn-iOS](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDAppAuthFetcherAuthorizationWithEMMSupport.m) that supported EMM error handling (with the `OIDAuthState` at hand). Thus, the `GIDGoogleUser` could potentially expose a [`fetcherAuthorizer`](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h#L71) that would capture the fetcher's callbacks and provide GSI with a chance to substitute a different (EMM) error if one occurs.

In this change, we got rid of the subclass because GTMAppAuth is now written fully in Swift, and `GTMAppAuthFetcherAuthorization` has been replaced with `GTMAuthSession`. That means `GTMAuthSession` cannot be subclassed in Objective-C. The solution we came up with was that GTMAppAuth would expose a new protocol [`GTMAuthSessionDelegate`](https://github.com/google/GTMAppAuth/blob/3.0.0/GTMAppAuth/Sources/AuthSessionDelegate.swift) that provided the various callbacks `GTMAppAuthFetcherAuthorization` previously provided, and also added an interface for EMM error handling. Since `GIDEmmSupport` already existed and did the error handling for this case, I just added the `GTMAuthSessionDelegate` conformance there.

Internally, [`GTMAuthSession` will check to see if it has a delegate](https://github.com/google/GTMAppAuth/blob/3.0.0/GTMAppAuth/Sources/AuthSession.swift#L290) and will ask it for an updated error when it is authorizing a request.

Putting it all together, if clients wish to use `GIDGoogleUser`'s `fetcherAuthorizer` to authorize their fetches, then they will get EMM error handling support.

Let me know if this makes sense!

GoogleSignIn/Sources/GIDAuthStateMigration.h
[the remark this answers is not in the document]
mdmathias: Fixed.

GoogleSignIn/Sources/GIDEMMSupport.m
```diff
@@ -44,8 +44,55 @@
...
+  ErrorCodeAppVerificationRequired,
+} ErrorCode;
+
 @implementation GIDEMMSupport
 
+- (instancetype)init {
+  return [super init];
+}
+
++ (nullable NSError *)handleTokenFetchEMMError:(nullable NSError *)error {
```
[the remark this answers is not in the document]
mdmathias: The original point of this method was just to update the error from the server with the EMM variation supplied by GSI. The other method that you ask about is used in various flows and presents error alerts in some cases.

In any case, the method I added isn't necessary and so I've removed it. Instead, I placed its implementation within `GIDEMMSupport`'s implementation of `-updatedErrorForAuthSession:originalError:` (which it implements to conform to `GTMAuthSessionDelegate`. I think it is more clear now. LMK.

I would like @XiangtianDai to review and verify this. :)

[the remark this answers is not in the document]
mdmathias: Actually, it looks like perhaps my approach is incomplete. In the current implementation, we have:

* [-[GIDAppAuthFetcherAuthorizationWithEMMSupport authorizeRequest:
       completionHandler:]](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDAppAuthFetcherAuthorizationWithEMMSupport.m#L116)
* Which calls [+[GIDEMMSupport handleTokenFetchError:completion:]](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDEMMSupport.m#L49)
* Which calls [+[GIDEMMErrorHandler handleErrorFromResponse:completion:]](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDEMMErrorHandler.m#L59)

`+[GIDEMMErrorHandler handleErrorFromResponse:completion:]` potentially displays some UI, which my approach does not do. My approach does replace the error and send it back up to GTMAppAuth to eventually hand back, but I don't do any UI.

@XiangtianDai please verify if I should be presenting some UI in this case. If so, I will need to find a way to do. It'll be somewhat tricky since the existing mechanism for that uses class methods and block callbacks.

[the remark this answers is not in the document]
XiangtianDai: Yes, it would be really nice if we can keep the existing behavior of showing the MDM error UI in token fetching that is not part of initial sign-in. If I understand the code correctly this PR shows the MDM error UI in the initial sign-in flow only. This does solve the majority of the use cases, as most apps only sign in once for user identity but don't keep accessing user's data on Google services. However, for these apps that do, and if the device stops being compliant per admin policy after the initial sign-in, without the MDM error UI the user would see access failures without understanding why. Of course, the user can always sign out and sign in again to get the error UI, but still that would be non-ideal user experience. 

Again, not absolutely necessary but really nice to have.

GoogleSignIn/Sources/GIDEMMSupport.m
```diff
@@ -44,8 +44,55 @@
 // New UIDevice system name for iOS.
 static NSString *const kNewIOSSystemName = @"iOS";
 
+// The error key in the server response.
+static NSString *const kErrorKey = @"error";
+
+// Error strings in the server response.
+static NSString *const kGeneralErrorPrefix = @"emm_";
```
XiangtianDai, line 51: These constants are repeated from GIDEMMErrorHandler -- would it be better to have the error code mapping logic in GIDEMMErrorHandler so this class is just a wrapper? This way we have cleaner separation of responsibilities. (Also related to the discussion in other other thread about showing error UI).

GoogleSignIn/Tests/Unit/GIDAuthStateMigrationTest.m
```diff
@@ -113,44 +122,49 @@ - (void)tearDown {
...
-      saveAuthorization:_mockGTMAppAuthFetcherAuthorization toKeychainForName:kKeychainName];
-
-  [GIDAuthStateMigration migrateIfNeededWithTokenURL:[NSURL URLWithString:kTokenURL]
-                                        callbackPath:kCallbackPath
-                                        keychainName:kKeychainName
-                                      isFreshInstall:NO];
+  [[[_mockUserDefaults expect] andReturnValue:@NO] boolForKey:kMigrationCheckPerformedKey];
+
+  NSError *keychainSaveError = [NSError new];
+  OCMStub([_mockGTMKeychainStore saveAuthSession:OCMOCK_ANY error:[OCMArg setTo:keychainSaveError]]);
```
XiangtianDai, line 158: Just curious on the plan for OCMock v3 or v2 style (https://ocmock.org/ocmock3/). Historically all tests in this repo are in OCMock v2 style, e.g., `[[mock stub] method];`. Do we want to start to have new code in OCMock v3 style, e.g., `OCMStub([mock method]);`, and potentially towards the goal of convert all tests to the new style, or shall we keep it consistent with OCMock v2 style?

GoogleSignIn/Sources/GIDEMMSupport.m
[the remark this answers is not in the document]
mdmathias: Sounds good. We will have to either refactor EMM error handling here or does a minor version release on GTMAppAuth to handle the `GIDGoogleUser` case.

Related, I think [GoogleSignIn-iOS 7.0.0 introduced a bug](https://github.com/google/GoogleSignIn-iOS/issues/304) where the `emmSupport` check will always fail, and will therefore not provide EMM error handling for the case we're discussing here.

We don't currently have any tests for this (though, I am adding something in this PR that will catch future regressions like this); @XiangtianDai are you able to test this somehow? Or, can you recommend a path for me?

GoogleSignIn/Tests/Unit/GIDAuthStateMigrationTest.m
mdmathias: I think the v3 style is better. Ultimately, our plan is to reduce and eventually remove mocking. Until then, we can gradually move over to the v3 style.

GoogleSignIn/Sources/GIDEMMSupport.m
mdmathias: Good idea; I've removed this duplicated constants from `GIDEMMSupport`.

GoogleSignIn/Sources/GIDEMMSupport.m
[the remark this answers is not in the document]
XiangtianDai: Yes, I can test the case manually once you're ready, just let me know. We still need unit test for this case though.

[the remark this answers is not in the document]
mdmathias: I believe I have already added a test for this in [-[GIDEMMSupportTest testEMMSupportDelegate]](https://github.com/google/GoogleSignIn-iOS/pull/299/files#diff-d51db5f699811af5b19c97475580c64f1c9bb2dd709c88041740e671b10e44db). Are you thinking of something different?

XiangtianDai (approved):

GoogleSignIn/Sources/GIDGoogleUser_Private.h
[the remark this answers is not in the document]
ch40w31: Thanks for the explanation. It's not as straightforward as it used to be now that `GIDEmmSupport` conforms to `GTMAuthSessionDelegate`, and `GIDAuthSession` conforms to `GTMAppAuthFetcherAuthorization`,  but I think this is a reasonable path moving forward. Resolving this conversation.

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDAuthStateMigration.m
```diff
@@ -55,14 +73,15 @@ + (void)migrateIfNeededWithTokenURL:(NSURL *)tokenURL
   // action and go on to mark the migration check as having been performed.
   if (!isFreshInstall) {
     // Attempt migration
-    GTMAppAuthFetcherAuthorization *authorization =
+    GTMAuthSession *authorization =
```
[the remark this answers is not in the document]
mdmathias: Changed throughout `GIDAuthStateMigration`.

ch40w31 (approved):

PR #301, opened 2023-04-19
Fix the restorePreviousSignIn when completion is nil

henryhl22321:
Return to void and not trigger the callback when the completion is nil.

### DIFF
--- a/GoogleSignIn/Tests/Unit/GIDSignInTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDSignInTest.m
@@ -503,6 +503,18 @@ static NSString *const kNewScope = @"newScope";
   [_authState verify];
 }
 
+- (void)testNotRestorePreviousSignWhenCompletionIsNil {
+  [[[_authorization expect] andReturn:_authState] authState];
+  [[[_authState expect] andReturnValue:[NSNumber numberWithBool:NO]] isAuthorized];
+  _completionCalled = NO;
+  _authError = nil;
+
+  [_signIn restorePreviousSignInWithCompletion:nil];
+  [NSThread sleepForTimeInterval:1.0];
+
+  XCTAssertNil(_signIn.currentUser);
+}
+
 - (void)testOAuthLogin {
   [self OAuthLoginWithAddScopesFlow:NO
                           authError:nil
--- a/GoogleSignIn/Tests/Unit/GIDSignInTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDSignInTest.m
@@ -516,24 +516,18 @@ static NSString *const kNewScope = @"newScope";
   [[[_authorization expect] andReturn:_authState] authState];
   [[[_authState expect] andReturnValue:[NSNumber numberWithBool:YES]] isAuthorized];
 
-  OIDAuthorizationResponse *authResponse =
-        [OIDAuthorizationResponse testInstanceWithAdditionalParameters:nil
-                                                           errorString:nil];
   OIDTokenResponse *tokenResponse =
-        [OIDTokenResponse testInstanceWithIDToken:[OIDTokenResponse fatIDToken]
-                                      accessToken:kAccessToken
-                                        expiresIn: nil
-                                     refreshToken:kRefreshToken
-                                     tokenRequest:nil];
+  [OIDTokenResponse testInstanceWithIDToken:[OIDTokenResponse fatIDToken]
+                                accessToken:kAccessToken
+                                  expiresIn:nil
+                               refreshToken:kRefreshToken
+                               tokenRequest:nil];
 
   [[[_authState stub] andReturn:tokenResponse] lastTokenResponse];
 
-  // SaveAuthCallback
-  __block OIDAuthState *authState;
-  __block GIDProfileData *profileData;
   [[[_user stub] andReturn:_user] alloc];
-        (void)[[[_user expect] andReturn:_user] initWithAuthState:SAVE_TO_ARG_BLOCK(authState)
-                                                      profileData:SAVE_TO_ARG_BLOCK(profileData)];
+  (void)[[[_user expect] andReturn:_user] initWithAuthState:OCMOCK_ANY
+                                                profileData:OCMOCK_ANY];
   XCTAssertNil(_signIn.currentUser);
 
   [_signIn restorePreviousSignInWithCompletion:nil];
@@ -1278,7 +1272,7 @@ static NSString *const kNewScope = @"newScope";
     [[[_authState expect] andReturn:tokenResponse] lastTokenResponse];
     if (oldAccessToken) {
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
-      // Corresponds to EMM support 
+      // Corresponds to EMM support
       [[[_authState expect] andReturn:authResponse] lastAuthorizationResponse];
 #endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
       [[[_authState expect] andReturn:tokenResponse] lastTokenResponse];
--- a/GoogleSignIn/Tests/Unit/GIDSignInTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDSignInTest.m
@@ -517,13 +517,18 @@ static NSString *const kNewScope = @"newScope";
   [[[_authState expect] andReturnValue:[NSNumber numberWithBool:YES]] isAuthorized];
 
   OIDTokenResponse *tokenResponse =
-  [OIDTokenResponse testInstanceWithIDToken:[OIDTokenResponse fatIDToken]
-                                accessToken:kAccessToken
-                                  expiresIn:nil
-                               refreshToken:kRefreshToken
-                               tokenRequest:nil];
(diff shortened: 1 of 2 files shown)


GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,18 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+  [[[_authorization expect] andReturn:_authState] authState];
+  [[[_authState expect] andReturnValue:[NSNumber numberWithBool:NO]] isAuthorized];
+  _completionCalled = NO;
+  _authError = nil;
+
+  [_signIn restorePreviousSignInWithCompletion:nil];
+  [NSThread sleepForTimeInterval:1.0];
+
+  XCTAssertNil(_signIn.currentUser);
+}
```
ch40w31, line 516: It appears that this test case mimics the previous `testRestorePreviousSignInWhenSignedOut` case except the callback is nil. The previous case tests that the user is "signed out", meaning that they don't have a stored token, therefore their sign in cannot be restored. IMO, it's more interesting to test the scenario that the user does have a stored token, and `restorePreviousSignInWithCompletion` succeeds with a nil callback. However, there isn't an existing standalone test case that tests the success scenario with a non-nil callback, so I think it's ok if you don't add a unit test in this PR.

ch40w31 (changes requested):

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,18 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
 }
 
+- (void)testNotRestorePreviousSignWhenCompletionIsNil {
+  [[[_authorization expect] andReturn:_authState] authState];
+  [[[_authState expect] andReturnValue:[NSNumber numberWithBool:NO]] isAuthorized];
+  _completionCalled = NO;
+  _authError = nil;
+
+  [_signIn restorePreviousSignInWithCompletion:nil];
+  [NSThread sleepForTimeInterval:1.0];
```
mdmathias, line 513: In general, it's not good practice to sleep. Also, I don't think this is needed here. I removed this line and ran the test repeatedly (100 times) and didn't have any failure.

You can control-click on the diamond next to a test and see the following below.

Screenshot 2023-04-20 at 11 07 19 AM

image

mdmathias (changes requested):

mdmathias:
@henryhl22321 you will need to rebase `main` here after https://github.com/google/GoogleSignIn-iOS/pull/302 is merged. #302 is needed so that the workflows pass.

mdmathias (changes requested):

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,45 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+  [_signIn restorePreviousSignInWithCompletion:nil];
+
+  XCTAssertNil(_signIn.currentUser);
+}
+
+- (void)testRestorePreviousSignInWhenCompletionIsNil {
+  [[[_authorization expect] andReturn:_authState] authState];
+  [[[_authState expect] andReturnValue:[NSNumber numberWithBool:YES]] isAuthorized];
+
+  OIDAuthorizationResponse *authResponse =
```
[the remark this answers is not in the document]
henryhl22321: Removed

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,45 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+        [OIDAuthorizationResponse testInstanceWithAdditionalParameters:nil
+                                                           errorString:nil];
+  OIDTokenResponse *tokenResponse =
+        [OIDTokenResponse testInstanceWithIDToken:[OIDTokenResponse fatIDToken]
+                                      accessToken:kAccessToken
+                                        expiresIn: nil
```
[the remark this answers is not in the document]
henryhl22321: Done

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,45 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+                                     refreshToken:kRefreshToken
+                                     tokenRequest:nil];
+
+  [[[_authState stub] andReturn:tokenResponse] lastTokenResponse];
+
+  // SaveAuthCallback
+  __block OIDAuthState *authState;
```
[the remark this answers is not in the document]
henryhl22321: Done

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,39 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+
+  XCTAssertNil(_signIn.currentUser);
+}
+
+- (void)testRestorePreviousSignInWhenCompletionIsNil {
+  [[[_authorization expect] andReturn:_authState] authState];
+  [[[_authState expect] andReturnValue:[NSNumber numberWithBool:YES]] isAuthorized];
+
+  OIDTokenResponse *tokenResponse =
+  [OIDTokenResponse testInstanceWithIDToken:[OIDTokenResponse fatIDToken]
```
mdmathias, line 520: Please fix the indentation to match the rest of the file. `[OIDTokenResponse ...` should be indented four spaces from the column where text begins on line 519.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,39 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+                                accessToken:kAccessToken
+                                  expiresIn:nil
+                               refreshToken:kRefreshToken
+                               tokenRequest:nil];
+
+  [[[_authState stub] andReturn:tokenResponse] lastTokenResponse];
+
+  [[[_user stub] andReturn:_user] alloc];
+  (void)[[[_user expect] andReturn:_user] initWithAuthState:OCMOCK_ANY
```
mdmathias, line 529: Did you decide to not capture the values of `authState` and `profileData` here to compare against what is on `_signIn.currentUser` below? I think it would add value to the test.

mdmathias (changes requested):

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -503,6 +503,46 @@ - (void)testRestorePreviousSignInWhenSignedOut {
...
+                                                profileData:OCMOCK_ANY];
+  XCTAssertNil(_signIn.currentUser);
+
+  [_signIn restorePreviousSignInWithCompletion:nil];
+
+  XCTAssertNotNil(_signIn.currentUser);
+  XCTAssertEqualObjects(_signIn.currentUser.userID, kUserID);
```
mdmathias, line 541: Testing equality with `kUserID` feels a little vague and makes it seem like `kUserID` is sort of magical. I can't read this test and say that I understand why `_signIn.currentUser.userID == kUserID`.

Is it possible to do something like:

```
__block OIDAuthState *capturedAuthState;
__block GIDProfileData *capturedProfileData;
(void)[[[_user expect] andReturn:_user] initWithAuthState:SAVE_TO_ARG_BLOCK(capturedAuthState)
                                                profileData:SAVE_TO_ARG_BLOCK(capturedProfileData)];
XCTAssertEqualObjects(_signIn.currentUser.authState, capturedAuthState);
XCTAssertEqualObjects(_signIn.currentUser.profileData, capturedProfileData);
```

If that's not easy or possible (probably due to mocking complexity), then let's talk about what individual asserts we can add here to give us confidence in the test. Additionally, you can add a comment with a link to an issue on this repo describing the problem, which you can use when working on removing all of these mocks in this test file.

henryhl22321: I did try the above way, but it throw the exception:
testRestorePreviousSignInWhenCompletionIsNil: ((_signIn.currentUser.profile) equal to (capturedProfileData)) failed: throwing "NSInternalInconsistencyException: OCClassMockObject(GIDGoogleUser): unexpected method invoked: profile 
	stubbed:	alloc"

GoogleSignIn/Tests/Unit/GIDSignInTest.m
mdmathias: Looks like it's too tricky for this PR to disentangle the mocking. #306 tracks the issue. 👍

GoogleSignIn/Tests/Unit/GIDSignInTest.m
henryhl22321: Created the issue #306 to track dropping the mock user, will be implemented in the future.

mdmathias (approved):

ch40w31 (approved):

PR #302, opened 2023-04-20
Remove macos-11 runner

mdmathias:
The macos-11 runner [fails to launch the test runner](https://github.com/google/GoogleSignIn-iOS/actions/runs/4694725464/jobs/8323148812), and so merging is blocked since tests are failing.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


ch40w31 (commented):
Should [this line](https://github.com/google/GoogleSignIn-iOS/blob/88f02926dae68923312bac97525fce294e267855/.github/workflows/builds.yml#L18) be updated too?

ch40w31 (approved):

PR #314, opened 2023-05-24
Add a prototype using Firebase App Check

mdmathias:
Adds support for using the Firebase App Check token in `GIDSignIn` sign in and add scopes methods. Also adds tests for some of these flow. 

I'm looking for general feedback on my approach and organization. This branch can either be merged into `mdmathias/app-attest` or we can wait until we have the final dependencies.

Up for discussion: whether or not we want to add tests for app check in `GIDSignInTest` for sign in and add scopes. We are actively working to make `GIDSignIn` more testable and remove mocking from GSI in general. Since we'd need to mock parts of `GIDSignIn` to test whether or not the app check token is added to the request, perhaps we'd like to wait to add those tests when we don't have to mock.


GoogleSignIn.podspec
```diff
@@ -13,7 +13,7 @@ The Google Sign-In SDK allows users to sign in with their Google account from th
     :tag => s.version.to_s
   }
   s.swift_version = '4.0'
-  ios_deployment_target = '10.0'
+  ios_deployment_target = '11.0'
```
paulb777, line 16: We may go to 12 in our fall major release since Xcode 15 drops iOS 11.

mdmathias: Are y'all considering that for Firebase as a whole, or just App Check? Either way, is that something we can discuss wrt App Check? I'd like to do some research on GSI usage by iOS version. A quick look suggests not many clients are on < iOS 12, but I'd like to be sure.

paulb777: We're planning to move Firebase as a whole to 13 except for leaving Analytics and Crashlytics at 11. Let's discuss what to do about AppCheck.

mdmathias: Following up here from an offline conversation. GSI is comfortable with an increase for the minimum iOS deployment target to 12.

GoogleSignIn/Sources/GIDSignIn_Private.h
```diff
@@ -44,10 +46,13 @@ typedef void (^GIDDisconnectCompletion)(NSError *_Nullable error);
 /// Redeclare |currentUser| as readwrite for internal use.
 @property(nonatomic, readwrite, nullable) GIDGoogleUser *currentUser;
 
-/// Private initializer for |GIDSignIn|.
-- (instancetype)initPrivate;
+/// Expose `GIDAppCheck` for tests.
```
ch40w31, line 49: We should avoid needing to expose a property for tests.

See the other comment regarding using composition for `id<GIDAppCheckProvider>` in `GIDSignIn` so you don't need to expose it here.

Samples/Swift/AppAttestExample/AppAttestExample.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
mdmathias: In the future, yes. Currently, my understanding is that this would be tricky because this sample project needs to be a Firebase project, which requires registration and a `GoogleService-Info.plist` file. In the future, we won't need this file, and we won't need the sample app to be a registered Firebase project (or rather, the developer will not need to do that on their own via the Firebase console). Therefore, I thought it'd be simpler to update our existing samples to exhibit this new functionality when we don't need a full Firebase project. We can delete this sample at that time.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDAppCheckProvider.h
```diff
@@ -0,0 +1,39 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class FIRAppCheckToken;
+
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+
+NS_AVAILABLE_IOS(14)
+@protocol GIDAppCheckProvider <NSObject>
```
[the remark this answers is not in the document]
mdmathias: Is there a reason why making this protocol is undesirable?

The trouble I have with using an initializer for composition is that `GIDSignIn` is currently a singleton, which makes using initializer composition tricky. Moreover, it would encourage and require writing tests that will depart from the actual API requirements of the class (i.e., unit tests would not go through the `GIDSignIn` shared singleton and would instead exercise a member-wise initializer that is only exposed internally.

Finally, this configuration method on `GIDSignIn` is using composition; admittedly, it's not initializer composition, but it is composition nonetheless.

```
- (void)configureWithAppCheckProvider:(nullable id<GIDAppCheckProvider>)provider
                           completion:(nullable void (^)(NSError * _Nullable error))completion
``` 

Last, I believe the `GIDAppCheck` does follow the pattern you describe: it has an initializer that takes an instance conforming to this protocol; if that instance is not nil, we will use that implementation for App Check work. Otherwise, we will use the conformance coming from Firebase (which we add ourselves elsewhere in this PR).

In summary, I think resolving this issue as you suggest would require a rewrite of `GIDSignIn`. Since that is beyond the scope of this PR, and is related to other ongoing work, I think we should hold off on these changes and pick them up when `GIDSignIn` is ready (i.e., when it is not a singleton).

Thoughts?

GoogleSignIn/Sources/GIDAppCheck.m
```diff
@@ -0,0 +1,107 @@
...
+    _appCheck = provider ?: [FIRAppCheck appCheck];
+    _workerQueue = dispatch_queue_create("com.google.googlesignin.appCheckWorkerQueue", nil);
+  }
+  return self;
+}
+
+- (void)prepareForAppCheckWithCompletion:(nullable void (^)(FIRAppCheckToken * _Nullable,
+                                                            NSError * _Nullable))completion {
+  dispatch_async(self.workerQueue, ^{
+    if (self.isPrepared) {
```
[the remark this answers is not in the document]
mdmathias: All reads and writes occur on `self.workerQueue`. Nonetheless, I can shadow the property internally and make them both `atomic`.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -66,6 +67,34 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
...
+- (void)configureWithAppCheckProvider:(nullable id<GIDAppCheckProvider>)provider
+                           completion:(nullable void (^)(NSError * _Nullable error))completion
+API_AVAILABLE(ios(14))
+NS_SWIFT_NAME(configureWithAppCheckProvider(_:completion:));
+
+/// Tells `GIDSignIn` to not request and use the Firebase App Check token.
+///
+/// Call this method, for example, when you do not wish to send the App Check token with the sign in
+/// or add scopes requests.
+- (void)turnOffAppCheck
```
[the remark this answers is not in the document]
mdmathias: My thinking was that developers may want a way to turn off passing in the app check token to the sign in or add scopes flows. Perhaps there was some error and they want to fall back?

GoogleSignIn/Sources/GIDSignIn_Private.h
mdmathias: We could put this property in the public header: `GIDSignIn.h`. What do you think? It's `readonly` so it shouldn't be problematic to expose.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDAppCheckProvider.h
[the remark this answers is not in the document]
ch40w31: As an SDK, exposing unnecessary types (or API surface in general) creates long-term maintenance burden. If we expose the `GIDAppCheckProvider` protocol, we may find app developers start to create their own fakes for their tests. And if we ever need to update the protocol, app developers would need to update their fakes, which is additional friction for them to use our latest SDK.

`GIDSignIn` already has a private initializer in GoogleSignIn/Sources/GIDSignIn_Private.h, you can add `id<GIDAppCheckProvider>` as an additional parameter to the initializer there.

We use this private initializer in our unit tests. The initialization of the shared singleton also calls through the private initializer so a decent level of consistency between tests and external use of the class is maintained.

Initializer composition and manual dependency injection through the initializer is a pattern we established in the `refactorGSI` branch. I think it's good to align our practices at this point so we don't need to refactor this code in refactorGSI again.

GoogleSignIn/Sources/GIDSignIn_Private.h
ch40w31: As a general principle, I'd prefer to keep the public API minimized as much as possible. A `readonly` property seems harmless, but once we make it public, app developers can use it in their ways (such as using it in tests as we do), and we're responsible for maintaining it. That would create additional hurdles for us if we ever need to refactor our implementation. Given that the App Check integration is very new, there's a great chance that we don't get it right the first time and find ourselves needing to refactor. I think it's particularly important to emphasize the need to minimize the public API at this point.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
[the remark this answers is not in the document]
mdmathias: I don't think having a field in the `Info.plist` won't handle the case that I have in mind: imagine a developer calls `-[GIDSignIn configureWithCompletion:` to prepare `GIDSignIn` to use App Check and everything goes fine. It works and `-[GIDSignIn useAppCheckToken]` is set to `YES`. That means sign in and add scopes calls will pass in the app check token. But maybe the developer doesn't want that for some reason.

Regardless, I hear what you're saying about the toggle's complexity. I will remove it for now.

GoogleSignIn/Sources/GIDSignIn_Private.h
mdmathias: Removed and reorganized file structure.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -0,0 +1,104 @@
...
+  if (self = [super init]) {
+    _prepared = NO;
+    _tokenFetcher = tokenFetcher ?: [FIRAppCheck appCheck];
+    _workerQueue = dispatch_queue_create("com.google.googlesignin.appCheckWorkerQueue", nil);
+  }
+  return self;
+}
+
+- (void)prepareForAppCheckWithCompletion:(nullable void (^)(FIRAppCheckToken * _Nullable,
+                                                            NSError * _Nullable))completion {
```
ch40w31, line 48: Can we prevent doing the following routine twice if there's a second call coming in before the first call returns?

The `prepared` bit isn't sufficient for that purpose, we'd need a bit like `preparationPending`.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -443,11 +456,43 @@ + (GIDSignIn *)sharedInstance {
...
+#pragma mark - Configuring and pre-warming
+
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable))completion {
+  @synchronized(self) {
+    // This check helps to avoid needing to set `_appCheck` in an initializer called in
+    // `+[GIDSignIn sharedInstance]`. Also defers to the fakes for `_appCheck` set in tests.
+    if (!_appCheck) {
+      _appCheck = [[GIDAppCheck alloc] initWithAppCheckTokenFetcher:nil];
+    }
```
ch40w31, line 475: IIUC, this is a temporary solution before we can put GIDAppCheck's initialization into the initializer of `GIDSignIn`. If so, can we leave a comment here to make it clear that this is temporary?

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -443,11 +456,43 @@ + (GIDSignIn *)sharedInstance {
...
+    [_appCheck prepareForAppCheckWithCompletion:^(FIRAppCheckToken * _Nullable token,
+                                                  NSError * _Nullable error) {
+      if (token) {
+        self->_useAppCheckToken = YES;
```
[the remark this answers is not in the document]
mdmathias: I think `self->` still is needed to suppress the warning below. Otherwise, we could do the weakify/strongify dance, but I was trying to match the style in the file.

image

GoogleSignIn/Sources/GIDSignIn.m
mdmathias: https://github.com/google/GoogleSignIn-iOS/issues/322

GoogleSignIn/Sources/GIDSignIn_Private.h
```diff
@@ -44,12 +45,19 @@ typedef void (^GIDDisconnectCompletion)(NSError *_Nullable error);
 /// Redeclare |currentUser| as readwrite for internal use.
 @property(nonatomic, readwrite, nullable) GIDGoogleUser *currentUser;
 
-/// Private initializer for |GIDSignIn|.
-- (instancetype)initPrivate;
+/// Whether or not the sign in or add scopes flows should use the Firebase App Check token
+@property(nonatomic, readonly) BOOL useAppCheckToken;
```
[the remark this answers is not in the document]
mdmathias: Agreed! Good point.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h
```diff
@@ -0,0 +1,46 @@
...
+#import <Foundation/Foundation.h>
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h"
+#import "GoogleSignIn/Sources/GIDAppCheck/API/GIDAppCheckProvider.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class FIRAppCheckToken;
+
+/// A list of potential error codes returned from the Google Sign-In SDK during App Check.
+typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDAppCheckErrorCode) {
```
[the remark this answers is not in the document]
mdmathias: I agree -- makes sense to have a separate error domain for `GIDAppCheck`. Also, good question about whether or not the app check errors will be exposed publicly. They will, since they can be passed thru via the completion called in `-[GIDSignIn configureWithCompletion:]`. I've added the error in a new file (`GIDAppCheckError.h`) that is public.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -0,0 +1,104 @@
...
+@property(nonatomic, strong) dispatch_queue_t workerQueue;
+@property(atomic, getter=isPrepared) BOOL prepared;
+
+@end
+
+@implementation GIDAppCheck
+
+- (instancetype)initWithAppCheckTokenFetcher:(nullable id<GIDAppCheckTokenFetcher>)tokenFetcher {
+  if (self = [super init]) {
+    _prepared = NO;
```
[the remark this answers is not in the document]
mdmathias: Good idea; added key storage via `NSUserDefaults`.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
mdmathias: I could see how that'd be helpful, but I can also see how that would be confusing. For example, what should we do if `preparing` is true? Do we call completion with an error?

```
- (void)prepareForAppCheckWithCompletion:(nullable void (^)(FIRAppCheckToken * _Nullable,
                                                            NSError * _Nullable))completion {
  if (self.preparing) {
    completion(nil, someErrorSayingThatPreparationIsAlreadyPending);
    return;
  }
  ...
}
```

I think the confusing part is that the caller, potentially at a different location than where the other pending prepare message was sent, will not receive a token or any indication of where to look for confirmation that preparation succeeded.

The way this is currently handled is to not guard against multiple inbound preparation requests, but to instead schedule those requests in order on the same async queue. This strategy doesn't prevent multiple, and wasteful, prepare requests, but it does avoid the confusion I mention above.

What do you think?

ch40w31: What you said makes sense. However, from the perspective of protecting the backend service that generates App Check tokens, we should make the API as safe as possible.

For multiple callbacks, a common technique is to save the callbacks in an array and invoked them altogether when the async operation is complete.

I think it's ok to leave a TODO for now and come back to it at a later point.

(Btw, do we need to return the token in the callback of this method?)

mdmathias: Fair enough. I've made a first pass at this in https://github.com/google/GoogleSignIn-iOS/pull/314/commits/68510d7fb22d304c306e298686771eb2036c8753. Let me know what you think. Happy to go thru feedback if you think it's the right approach. Otherwise, I can make an issue and put it there to address soon.

mdmathias: >  do we need to return the token in the callback of this method?

I'm a little torn. It is a little confusing to return the token here since this is just a prepare method; however, returning the token is a good signal of success. But perhaps we can remove the token parameter from completion and therefore the presence or not of an `NSError` would signal the same thing. What do you think?

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -23,16 +23,18 @@
 
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 
-/// Identifier for queue where App Check work is performed.
-static NSString *const kGIDAppCheckQueue = @"com.google.googlesignin.appCheckWorkerQueue";
 NSErrorDomain const kGIDAppCheckErrorDomain = @"com.google.GIDAppCheck";
 NSString *const kGIDAppCheckPreparedKey = @"com.google.GIDAppCheckPreparedKey";
 
+typedef void (^GIDAppCheckCompletion)(FIRAppCheckToken * _Nullable,NSError * _Nullable);
```
ch40w31, line 29: nit: add a space before `NSError *`

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -52,45 +56,64 @@ - (BOOL)isPrepared {
   return [self.userDefaults boolForKey:kGIDAppCheckPreparedKey];
 }
 
-- (void)prepareForAppCheckWithCompletion:(nullable void (^)(FIRAppCheckToken * _Nullable,
-                                                            NSError * _Nullable))completion {
+- (void)prepareForAppCheckWithCompletion:(nullable GIDAppCheckCompletion)completion {
+  if (completion) {
+    @synchronized (self.prepareCompletions) {
```
ch40w31, line 61: I think this should be `@synchronized (self)` instead because other accesses to `self.prepareCompletions` are synchronized on `self`.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -52,45 +56,64 @@ - (BOOL)isPrepared {
...
   dispatch_async(self.workerQueue, ^{
     if ([self isPrepared]) {
       NSError *error = [NSError errorWithDomain:kGIDAppCheckErrorDomain
                                            code:kGIDAppCheckAlreadyPrepared
                                        userInfo:nil];
-      if (completion) {
-        completion(nil, error);
+      @synchronized (self) {
+        for (GIDAppCheckCompletion savedCompletion in self.prepareCompletions) {
+          savedCompletion(nil, error);
```
ch40w31, line 81: Avoid calling the callback within `@synchornized` because the callback could be heavy and block other code in this class from entering `@synchornized`. You can make an immutable copy of the array in `@synchronized` and invoke the callbacks in the immutable copy outside `@synchronized`.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -52,45 +56,64 @@ - (BOOL)isPrepared {
...
-                                               code:kGIDAppCheckUnexpectedError
-                                           userInfo:nil];
-        completion(nil, noError);
+
+        for (GIDAppCheckCompletion savedCompletion in self.prepareCompletions) {
+          savedCompletion(token, maybeError);
+        }
+
+        [self.prepareCompletions removeAllObjects];
+        self.preparing = NO;
```
ch40w31, line 110: I think only access to `self.prepareCompletions` and `self.preparing` needs to be synchronzied.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
ch40w31: Given that we don't actually need a token at preparation time, I'd suggest we remove it. I think having just an `NSError` indicating success/failure of the preparation is pretty clear and self-explanatory.

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h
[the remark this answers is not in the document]
ch40w31: Let's discuss offline whether it makes sense to expose a new error domain publicly. Alternatively, we can add new codes to the existing `kGIDSignInErrorDomain`.

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
mdmathias: SGTM

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -557,19 +607,105 @@ - (void)signInWithOptions:(GIDSignInInternalOptions *)options {
...
+      shouldCallCompletion = NO;
+      GIDActivityIndicatorViewController *activityVC =
+          [[GIDActivityIndicatorViewController alloc] init];
+      [options.presentingViewController presentViewController:activityVC
+                                                     animated:true
+                                                   completion:^{
+        // Ensure that the activity indicator shows for at least 1/2 second to prevent "flashing"
+        dispatch_time_t halfSecond = dispatch_time(DISPATCH_TIME_NOW, NSEC_PER_MSEC / 2);
+        dispatch_after(halfSecond, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+          [self->_appCheck getLimitedUseTokenWithCompletion:
```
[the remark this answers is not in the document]
mdmathias: Would you be okay with keeping this way and adding a `TODO` with a link to an issue on this repo to replace this implementation with what you have outlined above?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -66,6 +67,22 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
...
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+
+/// Configures `GIDSignIn` for use.
+///
+/// @param completion A nullable callback block passing back any error arising from the
+/// configuration process if any exists.
+///
+/// Call this method on `GIDSignIn` prior to use and as early as possible. This method generates App
+/// Attest key IDs and the attestation object eagerly to minimize latency later on during the sign
+/// in or add scopes flows.
```
andrewheard, line 79: With regards to the potential error with code `kGIDAppCheckAlreadyPrepared` here, I think there are two scenarios:
1. The developer has called `configureWithCompletion:` twice in their code and the `kGIDAppCheckAlreadyPrepared` error is actionable (i.e., remove the extraneous configure call).
2. The app is being re-launched and is already prepared. Since developers are always (AFAICT) expected to call this method, I _think_ the completion block being called with an `kGIDAppCheckAlreadyPrepared` error may still be confusing (since they'd need to dig through the source code to see the comment that it's not a failing error). In this case the error isn't actionable and they should just ignore it.

Suggestion: IMHO, I think it would be ideal if the completion block was called with an error in scenario 1 but not scenario 2, potentially doing something like:
```
#if DEBUG
  NSLog(@"<INFO> App Check already prepared.");
#endif
```
Or not logging / emitting an error at all. Another alternative would be to always log a notice. I just worry that an `NSError` will be interpreted as an error, not a notice / info, regardless of the documentation around the `kGIDAppCheckAlreadyPrepared` error code.

Note: I didn't look into whether implementing my suggestion would be feasible -- just providing my two cents on what I think would be ideal.

mdmathias: I think we could implement your suggestion by having an api for debug, and an api for prod. That would be problematic though because that choice would impact developers.

I agree that there is some risk that developers will interpret this more seriously than we want, and having to look up the error's documentation is an extra hurdle. Perhaps the best way forward is to just log the "error" to the console.

I'll make the change but am also interested in @ch40w31's thoughts.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -443,11 +456,35 @@ + (GIDSignIn *)sharedInstance {
...
+#pragma mark - Configuring and pre-warming
+
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable))completion {
+  @synchronized(self) {
+    // This check helps to avoid needing to set `_appCheck` in an initializer called in
+    // `+[GIDSignIn sharedInstance]`. Also defers to the fakes for `_appCheck` set in tests.
+    // Ultimately, we would prefer to use `GIDSignIn`'s initializer to pass this information in, but
+    // this is currently impossible given that `GIDSignIn` is a singleton.
+    // TODO: Remove this once `GIDSignIn` is no longer a singleton (https://github.com/google/GoogleSignIn-iOS/issues/322)
```
ch40w31, line 475: I'm not sure this is accurate. Although `GIDSignIn` is a singleton at the public API level, it's been a non-singleton in our internal implementation and tests, because we can use the private initializer `initWithKeychainStore:`. What is blocking us from using `initWithKeychainStore:appCheckProvider:` for the singleton created in `sharedInstance` in this PR?

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDSignIn_Private.h
```diff
@@ -44,12 +45,16 @@ typedef void (^GIDDisconnectCompletion)(NSError *_Nullable error);
 /// Redeclare |currentUser| as readwrite for internal use.
 @property(nonatomic, readwrite, nullable) GIDGoogleUser *currentUser;
 
-/// Private initializer for |GIDSignIn|.
-- (instancetype)initPrivate;
-
-/// Private initializer taking a `GTMKeychainStore` to use during tests.
+/// Private initializer taking a `GTMKeychainStore` and `GIDAppCheckProvider` to use during tests.
```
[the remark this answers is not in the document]
mdmathias: Fixed the first. For the second, I can remove, but my understanding is that the reason why this method is exposed in  `GIDSignIn_Private.h` is to facilitate testing.

GoogleSignIn/Sources/GIDSignIn.m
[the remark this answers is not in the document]
mdmathias: https://github.com/google/GoogleSignIn-iOS/issues/329

GoogleSignIn/Sources/GIDSignIn.m
mdmathias: If we move creating `_appCheck` to `+[GIDSignIn sharedInstance]`, then we'll have something like this:

```
+ (GIDSignIn *)sharedInstance {
  static dispatch_once_t once;
  static GIDSignIn *sharedInstance;
  dispatch_once(&once, ^{
    GTMKeychainStore *keychainStore =
        [[GTMKeychainStore alloc] initWithItemName:kGTMAppAuthKeychainName];
    if (@available(iOS 14.0, *)) {
      sharedInstance = [[self alloc] initWithKeychainStore:keychainStore
                                          appCheckProvider:nil];
    }
#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
    if (!sharedInstance) {
      sharedInstance = [[self alloc] initWithKeychainStore:keychainStore];
    }
  });
  return sharedInstance;
}
```

This is fine, but then we'd need to remove `-[GIDSignInTest testSharedInstance]` since we are currently depending on Firebase proper, which requires that the `FirebaseApp` is configured prior to getting an instance of app check:

```
2023-07-18 16:38:16.089762-0700 xctest[2658:1509258] 10.10.0 - [FirebaseCore][I-COR000003] The default Firebase app has not yet been configured. Add `FirebaseApp.configure()` to your application initialization. This can be done in in the App Delegate's application(_:didFinishLaunchingWithOptions:)` (or the `@main` struct's initializer in SwiftUI). Read more: https://goo.gl/ctyzm8.
2023-07-18 16:38:16.092787-0700 xctest[2658:1508993] *** Terminating app due to uncaught exception 'com.firebase.appCheck', reason: 'The default FirebaseApp instance must be configured before the defaultAppCheck instance can be initialized. One way to ensure this is to call `FirebaseApp.configure()` in the App Delegate's `application(_:didFinishLaunchingWithOptions:)` (or the `@main` struct's initializer in SwiftUI).'
```

I've removed the singleton test since I don't think it does much more than test how singletons are implemented.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
mdmathias: I think I've captured the logging in scenario with this snippet in `-[GIDAppCheck prepareForAppCheckWithCompletion:]`

```
    if ([self isPrepared]) {
#if DEBUG
      NSLog(@"`GIDAppCheck` is already prepared.");
#endif
```

`[self isPrepared]` tracks whether or not configure has been called and has succeeded. We only log during `DEBUG` builds.

GoogleSignIn/Sources/GIDSignIn.m
ch40w31: Thanks! Can you add a TODO in `initWithKeychainStore:appCheckProvider:` to move initialization of `GIDAppCheck` to `sharedInstance` when Firebase App Check doesn't require `FirebaseApp.configure()`?

ch40w31 (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
mdmathias: Sorry, I think my reply was confusing. I believe I have addressed this already in [this commit](https://github.com/google/GoogleSignIn-iOS/pull/314/commits/2538f619ca9d3b7ffbbc450671623341f75c411d#diff-b76ad8011382daf05bf7a830db6abe2428c68fbeadd4b617b39bd2c504e2b640R535); however, to add further confusion, I have streamlined my approach in this [new commit](https://github.com/google/GoogleSignIn-iOS/pull/314/commits/74913be3d79b1a3a9f1661206b43ff9399cad136#diff-b76ad8011382daf05bf7a830db6abe2428c68fbeadd4b617b39bd2c504e2b640R463-R467).

mdmathias: I.e., no need for a TODO since we can create `GIDAppCheck` within `sharedInstance` if we do not have `-[GIDSignInTest testSharedInstance]`.

GoogleSignIn/Sources/GIDSignIn_Private.h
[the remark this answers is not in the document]
mdmathias: I have removed the text indicating that these methods are in place to facilitate testing.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
mdmathias: Per an offline conversation, I have [removed the `NSLog`](82b7ad8b4d6580965c3276a04dbf8e5adc6f3591) for previously in place for `DEBUG` builds.

ch40w31 (approved):

PR #328, opened 2023-07-17
Add HttpFetcher component

henryhl22321:
Add class HttpFetcher fetches data from URL endpoint. Add the FakeHttpFetcher.
Use HttpFetcher on GIDSignIn for revoking the grant scope.

### DIFF
--- a/GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
@@ -1,0 +1,106 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "GoogleSignIn/Sources/GIDHTTPFetcher/Implementations/GIDHTTPFetcher.h"
+
+#import "GoogleSignIn/Tests/Unit/OIDAuthState+Testing.h"
+
+#import <XCTest/XCTest.h>
+
+#ifdef SWIFT_PACKAGE
+@import GTMAppAuth;
+#else
+#import <GTMAppAuth/GTMAppAuth.h>
+#endif
+
+static NSString *const kTestURL = @"https://testURL.com";
+static NSString *const kErrorDomain = @"ERROR_DOMAIN";
+static NSInteger const kErrorCode = 400;
+
+@interface GIDHTTPFetcherTest : XCTestCase {
+  GIDHTTPFetcher *_httpFetcher;
+}
+
+@end
+
+@implementation GIDHTTPFetcherTest
+
+- (void)setUp {
+  [super setUp];
+  _httpFetcher = [[GIDHTTPFetcher alloc] init];
+}
+
+- (void)testFetchData_success {
+  NSURL *url = [NSURL URLWithString:kTestURL];
+  NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:url];
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GTMAuthSession *authorization = [[GTMAuthSession alloc] initWithAuthState:authState];
+  id<GTMSessionFetcherServiceProtocol> fetcherService = authorization.fetcherService;
+
+  GTMSessionFetcherTestBlock block =
+      ^(GTMSessionFetcher *fetcherToTest, GTMSessionFetcherTestResponse testResponse) {
+        NSData *data = [[NSData alloc] init];
+        testResponse(nil, data, nil);
+      };
+  [GTMSessionFetcher setGlobalTestBlock:block];
+
+  XCTestExpectation *expectation =
+      [self expectationWithDescription:@"Callback called with no error"];
+
+  [_httpFetcher fetchURLRequest:request
+                 withFetcherService:fetcherService
+                    withComment:@"Test data fetcher."
+                     completion:^(NSData *data, NSError *error) {
+    XCTAssertNil(error);
+    [expectation fulfill];
+  }];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
+}
+
+- (void)testFetchData_error {
+  NSURL *url = [NSURL URLWithString:kTestURL];
+  NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:url];
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GTMAuthSession *authorization = [[GTMAuthSession alloc] initWithAuthState:authState];
+  id<GTMSessionFetcherServiceProtocol> fetcherService = authorization.fetcherService;
+
+  GTMSessionFetcherTestBlock block =
+      ^(GTMSessionFetcher *fetcherToTest, GTMSessionFetcherTestResponse testResponse) {
+        NSData *data = [[NSData alloc] init];
+        NSError *error = [self error];
+        testResponse(nil, data, error);
+      };
+  [GTMSessionFetcher setGlobalTestBlock:block];
+
+  XCTestExpectation *expectation =
+      [self expectationWithDescription:@"Callback called with an error"];
+
+  [_httpFetcher fetchURLRequest:request
+             withFetcherService:fetcherService
+                    withComment:@"Test data fetcher."
+                     completion:^(NSData *data, NSError *error) {
+    XCTAssertNotNil(error);
+    XCTAssertEqual(error.code, kErrorCode);
+    [expectation fulfill];
+  }];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
+}
+
+#pragma mark - Helpers
+
+- (NSError *)error {
+  return [NSError errorWithDomain:kErrorDomain code:kErrorCode userInfo:nil];
+}
+
+@end
--- a/GoogleSignIn/Tests/Unit/GIDSignInTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDSignInTest.m
@@ -32,13 +32,13 @@
 #import "GoogleSignIn/Sources/GIDGoogleUser_Private.h"
 #import "GoogleSignIn/Sources/GIDSignIn_Private.h"
 #import "GoogleSignIn/Sources/GIDSignInPreferences.h"
+#import "GoogleSignIn/Sources/GIDHTTPFetcher/Implementations/Fakes/GIDFakeHTTPFetcher.h"
+#import "GoogleSignIn/Sources/GIDHTTPFetcher/Implementations/GIDHTTPFetcher.h"
 
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 #import "GoogleSignIn/Sources/GIDEMMErrorHandler.h"
 #endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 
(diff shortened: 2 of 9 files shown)


mdmathias:
Hey @henryhl22321 it looks like you have some [linting failiures](https://github.com/google/GoogleSignIn-iOS/actions/runs/5581508846/jobs/10199737675?pr=328):

```
- ERROR | xcodebuild:  /Users/runner/work/GoogleSignIn-iOS/GoogleSignIn-iOS/GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m:48:3: error: declaration of 'GTMSessionFetcherTestBlock' must be imported from module 'GTMSessionFetcher.GTMSessionFetcher' before it is required
```

You may be able to diagnose these failures more quickly by running `pod lib lint GoogleSignIn.podspec` locally before sending up your PR for review.

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GTMAuthSession *authorization = [[GTMAuthSession alloc] initWithAuthState:authState];
+  id<GTMSessionFetcherServiceProtocol> fetcherService = authorization.fetcherService;
+
+  GTMSessionFetcherTestBlock block =
+      ^(GTMSessionFetcher *fetcherToTest, GTMSessionFetcherTestResponse testResponse) {
+        NSData *data = [[NSData alloc] init];
+        testResponse(nil, data, nil);
+      };
+  [GTMSessionFetcher setGlobalTestBlock:block];
```
mdmathias, line 53: I'm a little confused about why this is required. Does the fake fetcher use a `GTMSessionFetcher`? If it does, could this be moved into the fake fetcher to keep the test clean? Or, should it remain here so that we can write a test around this block?

Not saying it isn't, but am looking for help to understand.

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+                 withFetcherService:fetcherService
+                    withComment:@"Test data fetcher."
+                     completion:^(NSData *data, NSError *error) {
+    XCTAssertNil(error);
+    [expectation fulfill];
+  }];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
+}
+
+- (void)testFetchData_error {
```
mdmathias, line 68: `testFetchDataError`

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+// limitations under the License.
+
+#import "GoogleSignIn/Sources/GIDHTTPFetcher/Implementations/GIDHTTPFetcher.h"
+
+#import "GoogleSignIn/Tests/Unit/OIDAuthState+Testing.h"
+
+#import <XCTest/XCTest.h>
+
+@import GTMAppAuth;
+
```
mdmathias, line 22: Remove line

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+  [GTMSessionFetcher setGlobalTestBlock:block];
+
+  XCTestExpectation *expectation =
+      [self expectationWithDescription:@"Callback called with no error"];
+
+  [_httpFetcher fetchURLRequest:request
+                 withFetcherService:fetcherService
+                    withComment:@"Test data fetcher."
+                     completion:^(NSData *data, NSError *error) {
+    XCTAssertNil(error);
```
mdmathias, line 62: Should probably add an assert that the `data` is not nil: `XCTAssertNotNil(data)`.

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+  [GTMSessionFetcher setGlobalTestBlock:block];
+
+  XCTestExpectation *expectation =
+      [self expectationWithDescription:@"Callback called with an error"];
+
+  [_httpFetcher fetchURLRequest:request
+             withFetcherService:fetcherService
+                    withComment:@"Test data fetcher."
+                     completion:^(NSData *data, NSError *error) {
+    XCTAssertNotNil(error);
```
mdmathias, line 90: Probably want to add `XCTAssertNil(data)`.

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+    XCTAssertEqual(error.code, kErrorCode);
+    [expectation fulfill];
+  }];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
```
mdmathias, line 94: Extract timeout to a constant.

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+- (void)testFetchData_error {
+  NSURL *url = [NSURL URLWithString:kTestURL];
+  NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:url];
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GTMAuthSession *authorization = [[GTMAuthSession alloc] initWithAuthState:authState];
+  id<GTMSessionFetcherServiceProtocol> fetcherService = authorization.fetcherService;
+
+  GTMSessionFetcherTestBlock block =
+      ^(GTMSessionFetcher *fetcherToTest, GTMSessionFetcherTestResponse testResponse) {
+        NSData *data = [[NSData alloc] init];
```
mdmathias, line 77: To make sure I understand, will this cause the `fetcherService` to pass back the `data` you create here? If so, do we expect there to be `data` when there is an `error`?

GoogleSignIn/Tests/Unit/GIDHTTPFetcherTest.m
```diff
@@ -0,0 +1,103 @@
...
+        NSError *error = [self error];
```
mdmathias, line 78: IMO, we could "dampen" this test by creating the error here (instead of using a method). That would make this test easier to read because it would reduce the number of other places a reader needs to go to understand what is being used within the test.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
 #pragma mark - Tests - disconnectWithCallback:
 
 // Verifies disconnect calls callback with no errors if access token is present.
-- (void)testDisconnect_accessToken {
+- (void)testDisconnect_accessTokenIsPresent {
```
mdmathias, line 879: I see; this is where you got the `_` in the test names for `GIDHTTPFetcherTest`. IMO, that underscore isn't needed as ObjC typically uses camel-case. I suggest we migrate to that style as we update existing tests and also use camel-case when we write new tests.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
...
   [[[_authorization expect] andReturn:_authState] authState];
   [[[_authState expect] andReturn:_tokenResponse] lastTokenResponse];
   [[[_tokenResponse expect] andReturn:kAccessToken] accessToken];
-  [[[_authorization expect] andReturn:_fetcherService] fetcherService];
```
mdmathias, line 876: Why remove this and add the below?

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
...
-  [_authorization verify];
-  [_authState verify];
-  [_tokenResponse verify];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
+  XCTAssertTrue(_keychainRemoved, @"should clear saved keychain name");
 }
 
 // Verifies disconnect if access token is present.
-- (void)testDisconnectNoCallback_accessToken {
+- (void)testDisconnectNoCallback_accessTokenIsPresent {
```
mdmathias, line 898: Camel-case

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
...
-  [_authorization verify];
-  [_authState verify];
-  [_tokenResponse verify];
+  [self waitForExpectationsWithTimeout:1 handler:nil];
+  XCTAssertTrue(_keychainRemoved, @"should clear saved keychain name");
 }
 
 // Verifies disconnect calls callback with no errors if refresh token is present.
-- (void)testDisconnect_refreshToken {
+- (void)testDisconnect_refreshTokenIsPresent {
```
mdmathias, line 912: Camel-case

GoogleSignIn/Tests/Unit/GIDSignInTest.m
mdmathias: Seems like you're doing this in each test you update. Let's make sure to update those as well if we make a change during PR review.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
...
-    if (error == nil) {
-      [accessTokenExpectation fulfill];
-    }
+    XCTAssertNil(error);
+    [accessTokenExpectation fulfill];
   }];
-  [self verifyAndRevokeToken:kAccessToken
-                 hasCallback:YES
-      waitingForExpectations:@[accessTokenExpectation]];
-  [_authorization verify];
```
mdmathias, line 887: Why are you removing the calls to `verify`? It looks like you're still using mocks in `GIDSignInTest`, but are removing expectations and verifications that what is expected is called. I think these changes run the risk of making the tests weaker: we are still mocking and and also removing some of the guardrails around the mocks.

Is there a way to no longer mock in these tests (here and throughout)? That is one of the main goals of this refactor.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -1230,41 +1222,35 @@ - (void)testTokenEndpointEMMError {
...
-  NSData *data = nil;
-  if (dataObject) {
-    NSError *jsonError = nil;
-    data = [NSJSONSerialization dataWithJSONObject:dataObject
-                                           options:0
-                                             error:&jsonError];
-    XCTAssertNil(jsonError, @"must provide valid data");
-  }
-  [_fetcherService.fetchers[0] didFinishWithData:data error:error];
+- (void)notDoFetch {
```
mdmathias, line 1239: Could this be renamed `doNotFectch`?

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -869,99 +876,89 @@ - (void)testNotHandleWrongPath {
...
 }
 
 // Verifies disconnect if access token is present.
-- (void)testDisconnectNoCallback_accessToken {
+- (void)testDisconnectNoCallback_accessTokenIsPresent {
   [[[_authorization expect] andReturn:_authState] authState];
   [[[_authState expect] andReturn:_tokenResponse] lastTokenResponse];
   [[[_tokenResponse expect] andReturn:kAccessToken] accessToken];
-  [[[_authorization expect] andReturn:_fetcherService] fetcherService];
+  OCMStub([_authorization fetcherService]).andReturn(_fetcherService);
```
mdmathias, line 902: I don't think you need to stub out the `fetcherService` on the `_authorization`. It is declared as a [`public var` on `GTMAuthSession`](https://github.com/google/GTMAppAuth/blob/master/GTMAppAuth/Sources/AuthSession.swift#L76). Is there a way that you can leverage that to reduce some of the mocking in these tests?

PR #331, opened 2023-08-01
Add a timed loader class

### LINKED ISSUE #329: Finalize and replace app check loading logic in -[GIDSignIn authorizationRequestWithOptions:completion:]`
The current implementation dispatches async after waiting 1/2 second to ensure that an activity spinner is displayed to avoid flashing. We should update the logic to instead do the following:

1. If the call to app check completes before some short threshold is met, just carry on without any loading screen.
2. If that threshold is exceeded, then display a loading screen for at least 1/2 second.
3. After displaying the loading screen for 1/2, complete when callback is received from app check.

mdmathias:
GIDTimedLoader encapsulates the logic to handle presenting an activity indicator when fetching the app check token.

Fixes #329

### DIFF
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -646,8 +646,8 @@ static NSString *const kConfigOpenIDRealmKey = @"GIDOpenIDRealm";
       if (_timedLoader.animationStatus != GIDTimedLoaderAnimationStatusAnimating) {
         [_timedLoader startTiming];
       }
-      [self->_appCheck getLimitedUseTokenWithCompletion:^(FIRAppCheckToken * _Nullable token,
-                                                          NSError * _Nullable error) {
+      [self->_appCheck getLimitedUseTokenWithCompletion:
+          ^(id<GACAppCheckTokenProtocol> _Nullable token, NSError * _Nullable error) {
         OIDAuthorizationRequest *request = nil;
         if (token) {
           additionalParameters[kClientAssertionTypeParameter] = kClientAssertionTypeParameterValue;
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -180,8 +180,10 @@ static NSString *const kConfigOpenIDRealmKey = @"GIDOpenIDRealm";
   BOOL _restarting;
   // Keychain manager for GTMAppAuth
   GTMKeychainStore *_keychainStore;
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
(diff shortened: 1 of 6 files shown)


GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -634,35 +639,29 @@ - (void)authorizationRequestWithOptions:(GIDSignInInternalOptions *)options comp
...
-            }
-            [activityVC.activityIndicator stopAnimating];
-            [activityVC dismissViewControllerAnimated:YES completion:nil];
-            completion(nil, error);
-            return;
+      UIViewController *presentingVC = options.presentingViewController;
+      if (!_timedLoader) {
+        _timedLoader = [[GIDTimedLoader alloc] initWithPresentingViewController:presentingVC];
+      }
+      if (_timedLoader.animationStatus != GIDTimedLoaderAnimationStatusAnimating) {
```
yalkhder, line 646: Do we need this check if GIDTimedLoader does that check anyway in the implementation of startTiming?

GoogleSignIn/Sources/GIDTimedLoader/GIDTimedLoader.h
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+#import <TargetConditionals.h>
+
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+
+typedef NS_ENUM(NSUInteger, GIDTimedLoaderAnimationStatus) {
```
yalkhder, line 22: Add declarative comments for enum.

GoogleSignIn/Sources/GIDTimedLoader/GIDTimedLoader.m
```diff
@@ -0,0 +1,118 @@
...
+                                              completion:nil];
+  });
+}
+
+- (void)stopTimingWithCompletion:(void (^)(void))completion {
+  if (self.animationStatus != GIDTimedLoaderAnimationStatusAnimating) {
+    return;
+  }
+
+  if (self.loadingTimer) {
```
yalkhder, line 83: The nil check is not needed here is it?

GoogleSignIn/Sources/GIDSignIn.m
mdmathias: Nope!

GoogleSignIn/Sources/GIDTimedLoader/GIDTimedLoader.m
mdmathias: Nope!

yalkhder (approved):

yalkhder (approved):

PR #335, opened 2023-08-24
Update GIDSignIn to handle placeholder app check tokens

mdmathias:


GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -632,15 +636,18 @@ - (void)authenticateInteractivelyWithOptions:(GIDSignInInternalOptions *)options
...
   if (@available(iOS 14.0, *)) {
     // Only use `_appCheck` (created via singleton `+[GIDSignIn sharedInstance]` call) if
-    // `-[GIDAppCheck prepareForAppCheckWithCompletion:]` has been called
-    if ([_appCheck isPrepared]) {
-      shouldCallCompletion = NO;
+    // `GIDAppCheck` has been successfully prepared OR if the developer has attempted to configure.
+    // If former is false and the latter true, then preparation step failed for some reason; we
+    // still want to try to pass along the app check token (it just may take longer since the
+    // pre-warm step failed).
+    if ([_appCheck isPrepared] || _configureAppCheckCalled) {
```
andrewheard, line 649: I _think_ this would be a breaking change for existing users. E.g., if a developer updates to a version of the SDK that includes this then `shouldCreateAuthRequest` would be `NO` unless they set up their app for App Attest (including adding the entitlement) or they call `configureWithCompletion:`.

mdmathias: If a developer upgrades to the version of GSI containing this code, but doesn't onboard to AppCheck and/or doesn't call the configure method, then:

* `shouldCreateAuthRequest` will start out as `YES` per line 639
* `[_appCheck isPrepared]` will be `NO`
* `_configureAppCheckCalled` will be `NO`
* `shouldCreateAuthRequest` will still be be `YES` because checks on 649 do not pass and `shouldCreateAuthRequest` will not be set to `NO` on line 650

Thus, this code will land on line 676 and enter the success block below:

```
if (shouldCreateAuthRequest) {
  OIDAuthorizationRequest *request = [self authorizationRequestWithOptions:options
                                                      additionalParameters:additionalParameters];
  completion(request, nil);
}
```

This will create the `OIDAuthorizationRequest` as before prior to this App Check work. Am I misunderstanding you or missing something?

andrewheard: Sorry, I somehow misread this as `!([_appCheck isPrepared] || _configureAppCheckCalled)`. Thanks for clarifying, this LGTM as is.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -652,9 +659,14 @@ - (void)authorizationRequestWithOptions:(GIDSignInInternalOptions *)options comp
...
           additionalParameters[kClientAssertionTypeParameter] = kClientAssertionTypeParameterValue;
           additionalParameters[kClientAssertionParameter] = token.token;
-          request = [self authorizationRequestWithOptions:options
-                                     additionalParameters:additionalParameters];
         }
+        #if DEBUG
+        if (error) {
+          NSLog(@"[Google Sign-In iOS]: Error retrieving App Check limited use token: %@", error);
+        }
+        #endif
```
andrewheard, line 667: Nice, I think this is a good compromise vs. entirely swallowing the error. In a follow-up, you could consider using [OSLog](https://developer.apple.com/documentation/os/logging/generating_log_messages_from_your_code?language=objc) like:
```
os_log_t customLog = os_log_create("com.google.sign-in", "App Check");
os_log_with_type(customLog, OS_LOG_TYPE_ERROR, Error retrieving App Check limited use token: %@", error);
```

Xcode 15 makes it easy to customize which logging levels, categories, etc. are shown if you use OSLog.

Samples/Swift/AppAttestExample/Podfile
```diff
@@ -1,10 +1,13 @@
...
+
 pod 'GoogleSignIn', :path => '../../../', :testspecs => ['unit']
 pod 'GoogleSignInSwiftSupport', :path => '../../../', :testspecs => ['unit']
 project 'AppAttestExample.xcodeproj'
 
 use_frameworks! :linkage => :static
 
 target 'AppAttestExample' do
-  pod 'AppCheckCore', :git => 'https://github.com/google/app-check.git', :tag => 'CocoaPods-0.1.0-alpha.1'
+  pod 'AppCheckCore'
```
andrewheard, line 11: Note: there will likely be more breaking changes coming (could use 'CocoaPods-0.1.0-alpha.6' or leave it as is to be quickly notified when AppCheckCore changes)

andrewheard (approved):

yalkhder (approved):

Samples/Swift/AppAttestExample/Podfile
mdmathias: Thanks for the heads up there. I think I'll it this way since I like learning about breaking changes sooner rather than later. We're just merging into a feature branch anyway...

GoogleSignIn/Sources/GIDSignIn.m
mdmathias: Thanks for the `OSLog` recommendation. I've created a tracking issue for this.

https://github.com/google/GoogleSignIn-iOS/issues/338

PR #336, opened 2023-08-25
Support using a debug app check provider during configuration

mdmathias:
Updates `-[GIDSignIn configureWithCompletion:]` to be `-[GIDSignIn configureDebugProviderWithAPIKey:completion:]` so that clients can specify using a debug app check provider (e.g., when running on the simulator).

### DIFF
--- a/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
+++ b/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
@@ -31,7 +31,7 @@ NS_CLASS_AVAILABLE_IOS(14)
 /// error.
 ///
 /// @param token The `GACAppCheckToken` instance to pass into the completion called from
-///     `getTokenWithCompletion:`. Use `nullable` if you would like a placeholder token from
+///     `getTokenWithCompletion:`. Use `nil` if you would like a placeholder token from
 ///     AppCheckCore.
 /// @param error The `NSError` to pass into the completion called from
 ///     `getTokenWithCompletion:`.
--- a/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
+++ b/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
@@ -31,7 +31,8 @@ NS_CLASS_AVAILABLE_IOS(14)
 /// error.
 ///
 /// @param token The `GACAppCheckToken` instance to pass into the completion called from
-///     `getTokenWithCompletion:`.
+///     `getTokenWithCompletion:`. Use `nullable` if you would like a placeholder token from
+///     AppCheckCore.
 /// @param error The `NSError` to pass into the completion called from
 ///     `getTokenWithCompletion:`.
 - (instancetype)initWithAppCheckToken:(nullable GACAppCheckToken *)token
--- a/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.m
+++ b/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.m
@@ -22,14 +22,14 @@ NSUInteger const kGIDAppCheckProviderFakeError = 1;
 
 @interface GIDAppCheckProviderFake ()
 
-@property(nonatomic, strong, nullable) id<GACAppCheckTokenProtocol> token;
+@property(nonatomic, strong, nullable) GACAppCheckToken *token;
 @property(nonatomic, strong, nullable) NSError *error;
 
 @end
 
 @implementation GIDAppCheckProviderFake
 
-- (instancetype)initWithAppCheckToken:(nullable id<GACAppCheckTokenProtocol>)token
+- (instancetype)initWithAppCheckToken:(nullable GACAppCheckToken *)token
                                 error:(nullable NSError *)error {
   if (self = [super init]) {
     _token = token;
@@ -38,14 +38,14 @@ NSUInteger const kGIDAppCheckProviderFakeError = 1;
   return self;
 }
 
-- (void)getTokenWithCompletion:(nonnull void (^)(id<GACAppCheckTokenProtocol> _Nullable,
-                                                 NSError * _Nullable))handler {
+- (void)getTokenWithCompletion:(void (^)(GACAppCheckToken * _Nullable,
+                                         NSError * _Nullable))handler {
   dispatch_async(dispatch_get_main_queue(), ^{
     handler(self.token, self.error);
   });
 }
 
-- (void)getLimitedUseTokenWithCompletion:(void (^)(id<GACAppCheckTokenProtocol> _Nullable,
+- (void)getLimitedUseTokenWithCompletion:(void (^)(GACAppCheckToken * _Nullable,
                                                    NSError * _Nullable))handler {
   dispatch_async(dispatch_get_main_queue(), ^{
     handler(self.token, self.error);
--- a/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.m
+++ b/GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.m
@@ -38,14 +38,13 @@ NSUInteger const kGIDAppCheckProviderFakeError = 1;
   return self;
 }
 
-- (void)getTokenWithCompletion:(void (^)(GACAppCheckToken * _Nullable,
-                                         NSError * _Nullable))handler {
+- (void)getTokenWithCompletion:(void (^)(GACAppCheckToken *, NSError * _Nullable))handler {
   dispatch_async(dispatch_get_main_queue(), ^{
     handler(self.token, self.error);
   });
 }
 
-- (void)getLimitedUseTokenWithCompletion:(void (^)(GACAppCheckToken * _Nullable,
+- (void)getLimitedUseTokenWithCompletion:(void (^)(GACAppCheckToken *,
                                                    NSError * _Nullable))handler {
   dispatch_async(dispatch_get_main_queue(), ^{
     handler(self.token, self.error);
--- a/GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h
+++ b/GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h
@@ -32,12 +32,15 @@ NS_CLASS_AVAILABLE_IOS(14)
 
 - (instancetype)init NS_UNAVAILABLE;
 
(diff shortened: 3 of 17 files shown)


GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h
```diff
@@ -30,13 +30,14 @@ extern NSString *const kGIDAppCheckPreparedKey;
...
 /// Creates the instance of this App Check wrapper class.
 ///
-/// The instance is created using `+[NSUserDefaults standardUserDefaults]` and the standard App
-/// Check provider.
+/// The instance is created using `+[NSUserDefaults standardUserDefaults]`.
 ///
-/// @SeeAlso The App Check provider is constructed with `+[GIDAppCheck standardAppCheckProvider]`.
-- (instancetype)init;
+/// @param useDebugProvider Whether or not the debug App Check provider should be used.
+- (instancetype)initWithDebugProvider:(BOOL)useDebugProvider;
```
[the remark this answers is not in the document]
mdmathias: I agree the name is misleading. I like your suggestion to use class methods.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -70,15 +70,17 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
...
 /// @param completion A nullable callback block passing back any error arising from the
-/// configuration process if any exists.
+///     configuration process if any exists.
 ///
 /// Call this method on `GIDSignIn` prior to use and as early as possible. This method generates App
 /// Attest key IDs and the attestation object eagerly to minimize latency later on during the sign
 /// in or add scopes flows.
-- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable error))completion
+- (void)configureWithDebugProvider:(BOOL)useDebugProvider
+                        completion:(nullable void (^)(NSError * _Nullable error))completion
```
andrewheard, line 81: I'm not sure if this would make the API too complex but have you considered having two methods like:
```
// Calls the method below with GIDAppCheckProviderAppAttest
- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable error))completion;
- (void)configureWithProvider:(GIDAppCheckProvider)appCheckProvider
                        completion:(nullable void (^)(NSError * _Nullable error))completion;
```

```
typedef NS_ENUM(NSInteger, GIDAppCheckProvider) {
    GIDAppCheckProviderAppAttest,
    GIDAppCheckProviderDebug
};
```

As a downside it would be a mistake if a developer called both configure methods. On the plus side it would allow additional attestation providers to be supported in the future. (E.g., despite being older, DeviceCheck support could be valuable if this work is extended to macOS in the future.)

andrewheard: If the above API is too complex, something else to consider would be:
```
- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable error))completion
               useDebugProvider:(BOOL)useDebugProvider
```
I don't like that the completion handler isn't at the end but I still worry that `configureWithDebugProvider` reads like you're configuring GSI to use the debug provider, despite it being a BOOL.

mdmathias: I did think about how to represent which debug provider to use. I like the enum idea, but given that the enum would only have two cases (and we don't have plans to extend this work to macOS just yet), an enum is essentially a `BOOL`. Additionally, I would like to bias against adding more types to the API surface since that increases what developers have to keep in their head and use, and also increases maintenance costs (admittedly only very incrementally for this change).

Finally, if we do decide to extend our support to DeviceCheck + macOS, then moving to an enum would be a great win with little refactoring required at the time. IMHO, we can wait for that time.

Samples/Swift/AppAttestExample/AppAttestExample/AppAttestExampleApp.swift
```diff
@@ -22,11 +22,19 @@ class AppDelegate: NSObject, UIApplicationDelegate {
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil
   ) -> Bool {
-    GIDSignIn.sharedInstance.configureWithCompletion { error in
+    #if DEBUG
```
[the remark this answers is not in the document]
mdmathias: Oh good idea; `targetEnvironment(simulator)` is better.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
andrewheard: nit: I still think `configureWithDebugProvider:` may be a confusing method name for developers. Without seeing the `BOOL` data type, I'd expect to be passing in an instance of a provider to that parameter.

I'm not sure if this is too verbose but what about:
```
- (void)configureWithDebugProviderEnabled:(BOOL)useDebugProvider
                        completion:(nullable void (^)(NSError * _Nullable error))completion
  NS_SWIFT_NAME(configure(debugProviderEnabled:completion:));
```

For Swift devs it'd look something like:
```
GIDSignIn.sharedInstance.configure(debugProviderEnabled: true) { error in
   // ...
}
```
`debugModeEnabled` or just `debugEnabled` might also work.

That said, I'm not super concerned since I think the current `NS_SWIFT_NAME(configure(usingDebugProvider:completion:))` reads reasonably well and 3P devs are probably more likely to be using the Swift API than the ObjC API.

andrewheard (approved):

yalkhder (approved):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
mdmathias: I agree. Updated.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -70,15 +70,17 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
...
 ///
 /// Call this method on `GIDSignIn` prior to use and as early as possible. This method generates App
 /// Attest key IDs and the attestation object eagerly to minimize latency later on during the sign
 /// in or add scopes flows.
-- (void)configureWithCompletion:(nullable void (^)(NSError * _Nullable error))completion
+- (void)configureWithDebugProviderEnabled:(BOOL)useDebugProvider
+                               completion:(nullable void (^)(NSError * _Nullable error))completion
 API_AVAILABLE(ios(14))
-NS_SWIFT_NAME(configureWithCompletion(completion:));
+NS_SWIFT_NAME(configure(debugProviderEnabled:completion:));
```
andrewheard, line 83: Now that we'll need an API key for the debug provider, but not for the standard App Attest flow, we may want to split this into two methods (`configureWithCompletion:` for default App Attest + a new one for debug).

cc: @weixifan

andrewheard (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
mdmathias: Good callout. What is here on GitHub is not up-to-date with what I have locally. I will push up that change soon.

GoogleSignIn/Tests/Unit/GIDAppCheckTest.m
```diff
@@ -57,9 +57,9 @@ - (void)testGetLimitedUseTokenFailure {
   GIDAppCheck *appCheck = [[GIDAppCheck alloc] initWithAppCheckProvider:fakeProvider
                                                            userDefaults:self.userDefaults];
 
-  [appCheck getLimitedUseTokenWithCompletion:^(id<GACAppCheckTokenProtocol> _Nullable token,
+  [appCheck getLimitedUseTokenWithCompletion:^(GACAppCheckToken * _Nullable token,
                                                NSError * _Nullable error) {
-    XCTAssertNil(token);
+    XCTAssertNotNil(token); // If there is an error, we expect a placeholder token
```
yalkhder, line 62: nit: How about also checking if the token is the placeholder token?

yalkhder (approved):

GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.m
```diff
@@ -35,7 +36,7 @@
 static NSString *const kGIDAppAttestBaseURL = @"https://firebaseappcheck.googleapis.com/v1beta";
 
 typedef void (^GIDAppCheckPrepareCompletion)(NSError * _Nullable);
-typedef void (^GIDAppCheckTokenCompletion)(id<GACAppCheckTokenProtocol> _Nullable,
+typedef void (^GIDAppCheckTokenCompletion)(GACAppCheckToken * _Nullable,
```
andrewheard, line 39: I think this also no longer needs to be `_Nullable`.

andrewheard (commented):
cc: @paulb777

GoogleSignIn/Tests/Unit/GIDAppCheckTest.m
mdmathias: I don't know what comparison that would be since the token will look and feel like a regular token, and it will be created and provided by AppCheckCore.

GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.m
```diff
@@ -22,14 +22,14 @@
 
 @interface GIDAppCheckProviderFake ()
 
-@property(nonatomic, strong, nullable) id<GACAppCheckTokenProtocol> token;
+@property(nonatomic, strong, nullable) GACAppCheckToken *token;
```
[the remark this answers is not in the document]
mdmathias: I've updated the signatures, but I would like this to be nullable since I will create instances of this fake with no token in scenarios where I want AppCheckCore to give me the placeholder token.

GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
```diff
@@ -31,7 +31,8 @@ NS_CLASS_AVAILABLE_IOS(14)
 /// error.
 ///
 /// @param token The `GACAppCheckToken` instance to pass into the completion called from
-///     `getTokenWithCompletion:`.
+///     `getTokenWithCompletion:`. Use `nullable` if you would like a placeholder token from
```
andrewheard, line 34: nit: "Use `nil` if you would like a placeholder token"

andrewheard (approved):

yalkhder (approved):

PR #337, opened 2023-08-31
Remove use of GACAppCheckTokenProtocol per AppCheckCore api changes

mdmathias:


Package.swift
```diff
@@ -48,7 +48,7 @@ let package = Package(
     .package(
       name: "AppCheck",
       url: "https://github.com/google/app-check.git",
-      .branch("main")),
+      .revision("023abd45d302912a4f8387de7ab2b741659faca6")),
```
andrewheard, line 51: It seems like you can specify a tag in place of a branch. This may be easier to track (if it works -- not tested):
```suggestion
      .branch("CocoaPods-0.1.0-alpha.6")),
```
Either way I'll try to keep you posted when there's a breaking change. It'll be easier when we tag the first release.

andrewheard (approved):

Package.swift
mdmathias: Good point regarding the tag. I think in this case that sha is for the release in question.

mdmathias: I'll change to the tag because it's more clear.

PR #339, opened 2023-09-12
Add SPM sample project for App Check

mdmathias:
This change adds an SPM sample project for App Check, updates the existing CocoaPods sample to derive from the SPM project so that they can stay in sync, adds a unit test reading the web API key from the bundle or the environment, does the same for the debug token, and sets up a workflow to pass in these values via environment variables.


Samples/Swift/AppAttestExample/README.md
```diff
@@ -0,0 +1,60 @@
...
+Do make sure that you do not commit this API key, or you will risk exposing
+this information on your repository.
+
+In builds running under continuous integration, make sure to use environment
+variables and `AppCheckSecretReader.swift` will find your web API key and debug
+token if you provide them to your `xcodebuild` command.
+
+## Integration Tests
+
+We should how you might hide your app's web API key and debug token when
```
yalkhder, line 53: Typo?

yalkhder (approved):

Samples/Swift/AppAttestExample/README.md
mdmathias: Ah yes. Thanks for catching.

Issue #346, opened 2023-11-02
GIDSignIn.sharedInstance.restorePreviousSignIn is not working as expected


darkForestCat:
**Describe the bug**

I've been using 'GoogleSignIn'  version '4.1.1' for many years and didn't have any trouble.Recently I've  updated to '7.0.0' because the older version does not support the Apple Silicon.Back on the '4.1.1' I've been using `GIDSignIn.sharedInstance().signInSilently()` to sign in the user that is already signed in.Then in the delegate method I was catching any kind of error.One of these errors is **_expired token_**.And on the older version everything was working perfect as a swiss watch while the newest version seems to have some troubles with catching errors **_on time_**.If I use '4.1.1' and I force sign out a Google account session with an app using this library everything works perfect, the delegate method handles immediately an error after the first `GIDSignIn.sharedInstance().signInSilently()` attempt, while the newest version where I'm forced to use this method:

```
GIDSignIn.sharedInstance.restorePreviousSignIn { user, error in    
            if error != nil || user == nil {
                self.service.authorizer = nil
            } else {
                self.service.authorizer = user?.fetcherAuthorizer
            }
 }
```

does not catch any kind of error and returns user without error as if everything is okay and no force sign out was done which leads to a bad user experience because app acts like if user is signed in, while actually not.The most interesting part is that after about 1 hour this method starts returning error saying "expired token", but again _**after waiting one hour**_.Does this method really supposed to give me an error about expired token after one hour and if so how to detect this error immediately without waiting so much time?

**To Reproduce**
1. Sign in inside your app.
2. Force sign out your app session from Devices section in the Security tab of your Google account settings.
3. Restart the app, the app will act like the token is not expired, unable to show you any kind of data.
4. Wait 1 hour but don't push Sign Out button in the app, just wait, after  1 hour `restorePreviousSignIn` method will return the error so the app will act like it should. 
**Even the test app has this issue, if you will sign out like in the step 2 and you will restart the test app it will show you that you are signed in...**

**Expected behavior**
I guess giving me the error immediately like the `sign(_ signIn: GIDSignIn!, didSignInFor user: GIDGoogleUser!, withError error: Error!)` method does in the version '4.1.1' without giving to a user any kind of bad experience.

**Environment**
- Device: [ iPhone XS Max]
- OS: [iOS 17.0 ]
- Pod version: [ 7.0.0]

[...]

PR #351, opened 2023-11-17
Clarifies that restored user is from local cache

camden-king:
This CL is in response to #346. To remove confusion I am adding to the `restorePreviousSignIn` docs that it will not refresh tokens if they are less than an hour old.

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
@@ -88,7 +88,7 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
 
 /// Attempts to restore a previous user sign-in without interaction. 
 ///
-/// Restores user from the local cache and refreshes tokens if they have expired (<1 hour).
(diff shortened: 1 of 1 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -86,8 +86,9 @@ typedef NS_ERROR_ENUM(kGIDSignInErrorDomain, GIDSignInErrorCode) {
 /// @return `YES` if there is a previous user sign-in saved in keychain.
 - (BOOL)hasPreviousSignIn;
 
-/// Attempts to restore a previous user sign-in without interaction. Uses cached user and doesn't refresh
-/// tokens if the tokens haven't expired (<1 hour).
+/// Attempts to restore a previous user sign-in without interaction. 
+///
+/// Uses cached user and doesn't refresh tokens if the tokens haven't expired (<1 hour).
```
mdmathias, line 91: I think this should explicitly state that the cache is local (on device). I also think the double negative after the "and" could be more clear.

`Restores user from the local cache and refreshes tokens if they have expired (<1 hour).`

PR #368, opened 2024-02-12
Check integration test for presubmit instruction

mdmathias:
Adds a new presubmit instruction who presence tells us to skip the integration test so that it's failure is not blocking.

SKIP_INTEGRATION_TESTS=YES


mdmathias:
Test comment for purely the purposes of testing.

yalkhder (approved):

PR #398, opened 2024-04-18
Implement GIDVerifyAccountDetail interface methods to start interactive flow. 

brnnmrls:
Implement `GIDVerifyAccountDetail` interface methods to start interactive flow by creating an instance of `GIDSignInInternalOptions` and calling a method to start the authorization flow passing in those options. Additionally, methods to generate a new configuration was moved from `GIDSignIn` to `GIDConfiguration`. 

1. Add new method, `-[GIDVerifyAccountDetail verifyAccountDetailsInteractivelyWithOptions:]`, that will start the authorization flow.
2. Implement `GIDVerifyAccountDetail` interface methods to generate a configuration from bundle.
3. Create a `GIDSignInInternalOptions` instance and call `-[GIDVerifyAccountDetail verifyAccountDetailsInteractivelyWithOptions:]`.

### DIFF
--- a/GoogleSignIn/Sources/GIDConfiguration.m
+++ b/GoogleSignIn/Sources/GIDConfiguration.m
@@ -26,6 +26,12 @@ static NSString *const kHostedDomainKey = @"hostedDomain";
 // The key for the openIDRealm property to be used with NSSecureCoding.
 static NSString *const kOpenIDRealmKey = @"openIDRealm";
 
+// Info.plist config keys
+static NSString *const kConfigClientIDKey = @"GIDClientID";
+static NSString *const kConfigServerClientIDKey = @"GIDServerClientID";
+static NSString *const kConfigHostedDomainKey = @"GIDHostedDomain";
+static NSString *const kConfigOpenIDRealmKey = @"GIDOpenIDRealm";
+
 NS_ASSUME_NONNULL_BEGIN
 
 @implementation GIDConfiguration
@@ -57,6 +63,35 @@ NS_ASSUME_NONNULL_BEGIN
     _openIDRealm = [openIDRealm copy];
   }
   return self;
+}
+
+// Try to retrieve a configuration value from an |NSBundle|'s Info.plist for a given key.
++ (nullable NSString *)configValueFromBundle:(NSBundle *)bundle forKey:(NSString *)key {
+  NSString *value;
+  id configValue = [bundle objectForInfoDictionaryKey:key];
+  if ([configValue isKindOfClass:[NSString class]]) {
+    value = configValue;
+  }
+  return value;
+}
+
++ (nullable instancetype)configurationFromBundle:(NSBundle *)bundle {
+  // Retrieve any valid config parameters from the bundle's Info.plist.
+  NSString *clientID = [self configValueFromBundle:bundle forKey:kConfigClientIDKey];
+  NSString *serverClientID = [self configValueFromBundle:bundle
+                                                  forKey:kConfigServerClientIDKey];
+  NSString *hostedDomain = [self configValueFromBundle:bundle forKey:kConfigHostedDomainKey];
+  NSString *openIDRealm = [self configValueFromBundle:bundle forKey:kConfigOpenIDRealmKey];
+
+  // If we have at least a client ID, try to construct a configuration.
+  if (clientID) {
+    return [[self alloc] initWithClientID:clientID
+                           serverClientID:serverClientID
+                             hostedDomain:hostedDomain
+                              openIDRealm:openIDRealm];
+  }
+
+  return nil;
 }
 
 // Extend NSObject's default description for easier debugging.
--- a/GoogleSignIn/Sources/GIDSignInInternalOptions.m
+++ b/GoogleSignIn/Sources/GIDSignInInternalOptions.m
@@ -36,6 +36,7 @@ NS_ASSUME_NONNULL_BEGIN
   GIDSignInInternalOptions *options = [[GIDSignInInternalOptions alloc] init];
   if (options) {
     options->_interactive = YES;
+    options->_continuation = NO; // Not relevant to VwG flow.
     options->_addScopesFlow = addScopesFlow;
     options->_configuration = configuration;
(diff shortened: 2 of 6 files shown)


GoogleSignIn/Sources/GIDConfiguration.m
```diff
@@ -26,6 +26,12 @@
 // The key for the openIDRealm property to be used with NSSecureCoding.
 static NSString *const kOpenIDRealmKey = @"openIDRealm";
 
+// Info.plist config keys
+static NSString *const kConfigClientIDKey = @"GIDClientID";
```
mdmathias, line 30: I think these are also defined on [GIDSignIn.m](https://github.com/google/GoogleSignIn-iOS/blob/ec20caf39917e3a50905776db1713b37c6b33840/GoogleSignIn/Sources/GIDSignIn.m#L139), right? Perhaps we can make a constant types so that we don't have duplicate code?

mdmathias: Or, if they're no longer used in `GIDSignIn`, then maybe just remove them from there?

GoogleSignIn/Sources/GIDSignInInternalOptions.m
```diff
@@ -36,7 +36,6 @@ + (instancetype)defaultOptionsWithConfiguration:(nullable GIDConfiguration *)con
   GIDSignInInternalOptions *options = [[GIDSignInInternalOptions alloc] init];
   if (options) {
     options->_interactive = YES;
-    options->_continuation = NO;
```
mdmathias, line 39: Why is this getting removed?

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,24 +31,39 @@ - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDet
...
+                  completion:completion];
 }
 
 - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDetails
     presentingViewController:(UIViewController *)presentingViewController
                         hint:(nullable NSString *)hint
                   completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
                                                 NSError *_Nullable error))completion {
-  // TODO(#383): Implement this method.
+  // Get the bundle of the current executable.
```
mdmathias, line 45: Don't need this comment.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,24 +31,39 @@ - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDet
...
+  NSBundle *bundle = NSBundle.mainBundle;
+
+  // If we have a bundle, try to set the active configuration from the bundle's Info.plist.
```
mdmathias, line 48: Same: remove comment.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,24 +31,39 @@ - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDet
...
 }
 
-- (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDetails
-    presentingViewController:(UIViewController *)presentingViewController
-                        hint:(nullable NSString *)hint
-            additionalScopes:(nullable NSArray<NSString *> *)additionalScopes
-                  completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
-                                                NSError *_Nullable error))completion {
-  // TODO(#383): Implement this method.
+// Starts authorization flow using the provided options.
```
mdmathias, line 64: This comment isn't necessary. It's duplicative of the doc comment in the header file.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDConfiguration.m
brnnmrls: Removed since they're no longer used in `GIDSignIn`.

GoogleSignIn/Sources/GIDSignInInternalOptions.m
brnnmrls: I removed this since we won't have to worry about the "continuation" property for the VwG flow and will only be used for the EMM flow. I thought since we won't make use of this in our flow then it wouldn't need to be filled. Should 'NO' still be passed to be safe?

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDSignInInternalOptions.m
brnnmrls: Sgtm, done!

mdmathias (approved):

PR #399, opened 2024-04-23
Depend on GSI 7.1.0 in GoogleSignSwiftSupport.podspec

mdmathias:
This change makes it so that `GoogleSignInSwiftSupport.podspec` depends upon v7.1.0 (and up) on CocoaPods instead of `"7.1.0-fac-beta-1.1.0`". This helps to avoid needing to release a new version of `GoogleSignInSwiftSupport` on CocoaPods, which would be required if we changed `GoogleSignInSwiftSupport`'s GSI dependency version.

SKIP_INTEGRATION_TESTS=YES

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -20,9 +20,10 @@ jobs:
           "",
           "--use-static-frameworks"
         ]
-        include:
-          - podspec: GoogleSignInSwiftSupport.podspec
-            includePodspecFlag: "--include-podspecs='GoogleSignIn.podspec'"
+        # See #400 (https://github.com/google/GoogleSignIn-iOS/issues/400)
+        # include:
+        #  - podspec: GoogleSignInSwiftSupport.podspec
+        #    includePodspecFlag: "--include-podspecs='GoogleSignIn.podspec'"
     steps:
     - uses: actions/checkout@v3
     - name: Update Bundler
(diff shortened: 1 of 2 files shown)


mdmathias:
`GoogleSignInSwiftSupport.podspec` fails to lint because that podspec lists its dependency on GSI as `s.dependency 'GoogleSignIn', '~> 7.1'`. Yet, the lint command is 

```
pod lib lint GoogleSignInSwiftSupport.podspec --verbose --no-clean \
     --sources=https://cdn.cocoapods.org/ \
     --include-podspecs='GoogleSignIn.podspec'
```

This will try to use the local `GoogleSignIn.podspec` for the Swift Support's podspec dependency. GSI's version in the local podspec is `'7.1.0-fac-beta-1.1.0'`. Thus, the linting fails with the error: 

```
-> GoogleSignInSwiftSupport (7.1.0)
    - ERROR | [iOS] unknown: Encountered an unknown error (CocoaPods could not find compatible versions for pod "GoogleSignIn":
  In Podfile:
    GoogleSignIn (from `.`)
```

Essentially, the lint command is saying to use the local `GoogleSignIn.podspec` to satisfy `GoogleSignInSwiftSupport.podspec`'s dependency on GSI, but the linting fails because the versions do not match.

If I remove using the local `GoogleSignIn.podspec` from the lint command, then linting will use GSI v7.1.0 (published on cocoapods.org) to satisfy `GoogleSignInSwiftSupport.podspec`'s dependency. The command below passes locally:

```
pod lib lint GoogleSignInSwiftSupport.podspec --verbose \
     --sources=https://cdn.cocoapods.org
```

Thus, I think this is fine to merge as is. We want `GoogleSignInSwiftSupport.podspec` to depend upon GSI v7.1.0 to avoid making a new release for the Swift Support pod. Additionally, `"7.1.0-fac-beta-1.1.0"` will be a published version of GSI upon release. This means future `pod lib lint`s won't fail.

mdmathias:
Actually, I've changed my mind! Disregard most of https://github.com/google/GoogleSignIn-iOS/pull/399#issuecomment-2071347783 in favor of #400.

yalkhder (approved):

PR #401, opened 2024-04-25
Add sanity checks for passed in options and initializers to GIDVerifyAccountDetail.

brnnmrls:
Begin implementing `-[GIDVerifyAccountDetail verifyAccountDetailsInteractivelyWithOptions:]` starting with input validation to ensure that the client has passed in valid parameters, provided a valid presenting view controller, and defined the proper callback schemes. Additionally, add tests for `GIDVerifyAccountDetail` like ensuring that a configuration was set which will be covered by the created initializers.

### DIFF
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -17,7 +17,6 @@
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h"
 
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDConfiguration.h"
-#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDConfiguration.h"
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiableAccountDetail.h"
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiedAccountDetailResult.h"
 
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -27,7 +27,7 @@
 
 @implementation GIDVerifyAccountDetail
 
-- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
+- (nullable instancetype)initWithConfig:(GIDConfiguration *)configuration {
   self = [super init];
   if (self) {
     _configuration = configuration;
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -27,7 +27,7 @@
 
 @implementation GIDVerifyAccountDetail
 
-- (instancetype)initWithConfig:(nullable GIDConfiguration *)configuration {
+- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
   self = [super init];
   if (self) {
     _configuration = configuration;
@@ -35,11 +35,15 @@
   return self;
 }
 
-- (instancetype)init {
+- (nullable instancetype)init {
   GIDConfiguration *configuration;
   NSBundle *bundle = NSBundle.mainBundle;
   if (bundle) {
     configuration = [GIDConfiguration configurationFromBundle:bundle];
+  }
+
+  if (!configuration) {
+    return nil;
   }
 
   return [self initWithConfig:configuration];
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -16,8 +16,12 @@
 
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h"
 
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDConfiguration.h"
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiableAccountDetail.h"
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiedAccountDetailResult.h"
+
+#import "GoogleSignIn/Sources/GIDSignInInternalOptions.h"
+#import "GoogleSignIn/Sources/GIDSignInCallbackSchemes.h"
 
 #if TARGET_OS_IOS
 
@@ -45,6 +49,59 @@
                   completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
                                                 NSError *_Nullable error))completion {
   // TODO(#383): Implement this method.
+}
+
+// Starts authorization flow using the provided options.
+- (void)verifyAccountDetailsInteractivelyWithOptions:(GIDSignInInternalOptions *)options {
+  if (!options.interactive) {
+    return;
+  }
+  
+  // Ensure that a configuration is set.
+  if (!_configuration) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"No active configuration. Make sure GIDClientID is set in Info.plist."];
+    return;    
+  }
+  
+  // Explicitly throw exception for missing client ID here. This must come before
+  // scheme check because schemes rely on reverse client IDs.
+  [self assertValidParameters:options];
+  
+  [self assertValidPresentingViewController:options];
+  
+  // If the application does not support the required URL schemes tell the developer so.
+  GIDSignInCallbackSchemes *schemes =
+  [[GIDSignInCallbackSchemes alloc] initWithClientIdentifier:options.configuration.clientID];
+  NSArray<NSString *> *unsupportedSchemes = [schemes unsupportedSchemes];
+  if (unsupportedSchemes.count != 0) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"Your app is missing support for the following URL schemes: %@",
+     [unsupportedSchemes componentsJoinedByString:@", "]];
+  }
+  // TODO(#397): Start the incremental authorization flow.
+}
+
+#pragma mark - Helpers
+
+// Asserts the parameters being valid.
+- (void)assertValidParameters:(GIDSignInInternalOptions *)options {
+  if (![options.configuration.clientID length]) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"You must specify |clientID| in |GIDConfiguration|"];
+  }
+}
+
+// Assert that the presenting view controller has been set.
+- (void)assertValidPresentingViewController:(GIDSignInInternalOptions *)options {
+  if (!options.presentingViewController) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"|presentingViewController| must be set."];
+  }
 }
 
 @end
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -27,7 +27,7 @@
 
 @implementation GIDVerifyAccountDetail
 
-- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
+- (instancetype)initWithConfig:(nullable GIDConfiguration *)configuration {
   self = [super init];
   if (self) {
     _configuration = configuration;
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -38,10 +38,7 @@
 
 - (instancetype)init {
   GIDConfiguration *configuration;
-  // Get the bundle of the current executable.
   NSBundle *bundle = NSBundle.mainBundle;
-
-  // If we have a bundle, try to set the active configuration from the bundle's Info.plist.
   if (bundle) {
     configuration = [GIDConfiguration configurationFromBundle:bundle];
   }
@@ -77,26 +74,25 @@
   [self verifyAccountDetailsInteractivelyWithOptions:options];
 }
 
-// Starts authorization flow using the provided options.
 - (void)verifyAccountDetailsInteractivelyWithOptions:(GIDSignInInternalOptions *)options {
   if (!options.interactive) {
     return;
   }
-  
+
   // Ensure that a configuration is set.
   if (!_configuration) {
     // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
     [NSException raise:NSInvalidArgumentException
                 format:@"No active configuration. Make sure GIDClientID is set in Info.plist."];
-    return;    
+    return;
   }
-  
+
   // Explicitly throw exception for missing client ID here. This must come before
   // scheme check because schemes rely on reverse client IDs.
   [self assertValidParameters:options];
-  
+
   [self assertValidPresentingViewController:options];
-  
+
   // If the application does not support the required URL schemes tell the developer so.
   GIDSignInCallbackSchemes *schemes =
   [[GIDSignInCallbackSchemes alloc] initWithClientIdentifier:options.configuration.clientID];
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -27,7 +27,7 @@
 
 @implementation GIDVerifyAccountDetail
 
-- (nullable instancetype)initWithConfig:(GIDConfiguration *)configuration {
+- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
   self = [super init];
   if (self) {
     _configuration = configuration;
@@ -35,7 +35,7 @@
   return self;
 }
 
-- (nullable instancetype)init {
+- (instancetype)init {
   GIDConfiguration *configuration;
   NSBundle *bundle = NSBundle.mainBundle;
   if (bundle) {
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -63,7 +63,55 @@
 
 // Starts authorization flow using the provided options.
 - (void)verifyAccountDetailsInteractivelyWithOptions:(GIDSignInInternalOptions *)options {
-  // TODO(#397): Sanity checks and start the incremental authorization flow.
+  if (!options.interactive) {
+    return;
+  }
+  
+  // Ensure that a configuration is set.
+  if (!_configuration) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"No active configuration. Make sure GIDClientID is set in Info.plist."];
+    return;    
+  }
+  
+  // Explicitly throw exception for missing client ID here. This must come before
+  // scheme check because schemes rely on reverse client IDs.
+  [self assertValidParameters:options];
+  
+  [self assertValidPresentingViewController:options];
+  
+  // If the application does not support the required URL schemes tell the developer so.
+  GIDSignInCallbackSchemes *schemes =
+  [[GIDSignInCallbackSchemes alloc] initWithClientIdentifier:options.configuration.clientID];
+  NSArray<NSString *> *unsupportedSchemes = [schemes unsupportedSchemes];
+  if (unsupportedSchemes.count != 0) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"Your app is missing support for the following URL schemes: %@",
+     [unsupportedSchemes componentsJoinedByString:@", "]];
+  }
+  // TODO(#397): Start the incremental authorization flow.
+}
+
+#pragma mark - Helpers
+
+// Asserts the parameters being valid.
+- (void)assertValidParameters:(GIDSignInInternalOptions *)options {
+  if (![options.configuration.clientID length]) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"You must specify |clientID| in |GIDConfiguration|"];
+  }
+}
+
+// Assert that the presenting view controller has been set.
+- (void)assertValidPresentingViewController:(GIDSignInInternalOptions *)options {
+  if (!options.presentingViewController) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"|presentingViewController| must be set."];
+  }
 }
 
 @end
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -27,6 +27,29 @@
 
 @implementation GIDVerifyAccountDetail
 
+- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
+  self = [super init];
+  if (self) {
+    _configuration = configuration;
+  }
+  return self;
+}
+
+- (instancetype)init {
+  GIDConfiguration *configuration;
+  // Get the bundle of the current executable.
+  NSBundle *bundle = NSBundle.mainBundle;
+
+  // If we have a bundle, try to set the active configuration from the bundle's Info.plist.
+  if (bundle) {
+    configuration = [GIDConfiguration configurationFromBundle:bundle];
+  }
+
+  return [self initWithConfig:configuration];
+}
+
+#pragma mark - Public methods
+
 - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDetails
     presentingViewController:(UIViewController *)presentingViewController
                   completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
@@ -42,14 +65,6 @@
                         hint:(nullable NSString *)hint
                   completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
                                                 NSError *_Nullable error))completion {
-  // Get the bundle of the current executable.
-  NSBundle *bundle = NSBundle.mainBundle;
-
-  // If we have a bundle, try to set the active configuration from the bundle's Info.plist.
-  if (bundle) {
-    _configuration = [GIDConfiguration configurationFromBundle:bundle];
-  }
-
   GIDSignInInternalOptions *options =
   [GIDSignInInternalOptions defaultOptionsWithConfiguration:_configuration
                                    presentingViewController:presentingViewController
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -88,6 +88,8 @@
     return;
   }
 
+  [self assertValidCurrentUser];
+
   // Explicitly throw exception for missing client ID here. This must come before
   // scheme check because schemes rely on reverse client IDs.
   [self assertValidParameters:options];
@@ -108,6 +110,15 @@
 }
 
 #pragma mark - Helpers
+
+// Assert that a current user exists.
+- (void)assertValidCurrentUser {
+  if (!GIDSignIn.sharedInstance.currentUser) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"|currentUser| must be set to verify."];
+  }
+}
 
 // Asserts the parameters being valid.
 - (void)assertValidParameters:(GIDSignInInternalOptions *)options {
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -74,6 +74,8 @@
   [self verifyAccountDetailsInteractivelyWithOptions:options];
 }
 
+#pragma mark - Authentication flow
+
 - (void)verifyAccountDetailsInteractivelyWithOptions:(GIDSignInInternalOptions *)options {
   if (!options.interactive) {
     return;
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
@@ -31,7 +31,10 @@
     presentingViewController:(UIViewController *)presentingViewController
                   completion:(nullable void (^)(GIDVerifiedAccountDetailResult *_Nullable verifyResult,
                                                 NSError *_Nullable error))completion {
-  // TODO(#383): Implement this method.
+  [self verifyAccountDetails:accountDetails
+    presentingViewController:presentingViewController
+                        hint:nil
+                  completion:completion];
 }
 
 - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDetails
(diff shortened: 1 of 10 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -43,6 +43,9 @@ typedef void (^GIDVerifyCompletion)(GIDVerifiedAccountDetailResult *_Nullable ve
 /// The active configuration for this instance of `GIDVerifyAccountDetail`.
 @property(nonatomic, nullable) GIDConfiguration *configuration;
 
+- (instancetype)initWithConfig:(GIDConfiguration *)config
```
mdmathias, line 46: Seems like this can be `nil` from the `.m` file. Mark as `nullable`.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,6 +27,26 @@
 
 @implementation GIDVerifyAccountDetail
 
+- (instancetype)initWithConfig:(GIDConfiguration *)configuration {
```
mdmathias, line 30: `GIDConfiguration` should be marked `nullable`.

mdmathias, line 30: Please add a test for this.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+@implementation GIDVerifyAccountDetailTests
+
+#pragma mark - Lifecycle
+
+- (void)setUp {
+  [super setUp];
+
+  _presentingViewController = [[UIViewController alloc] init];
+
+  _verifyAccountDetail = [[GIDVerifyAccountDetail alloc] init];
```
mdmathias, line 51: You should add variants of all tests using a passed `GIDConfiguration`.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+
+  GIDVerifiableAccountDetail *ageOver18Detail = [[GIDVerifiableAccountDetail alloc] initWithAccountDetailType:GIDAccountDetailTypeAgeOver18];
+  _verifiableAccountDetails = @[ageOver18Detail];
+
+  _fakeMainBundle = [[GIDFakeMainBundle alloc] init];
+  [_fakeMainBundle startFakingWithClientID:kClientId];
```
mdmathias, line 57: Only start faking in tests that will need this. E.g., the tests for flows using a passed `GIDConfiguration` will not need to fake the bundle.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+  [_fakeMainBundle fakeAllSchemesSupported];
```
mdmathias, line 58: Same here. Only fake in methods that need to fake.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+    threw = YES;
+    XCTAssertEqualObjects(exception.description,
+                          @"You must specify |clientID| in |GIDConfiguration|");
+  } @finally {
+  }
+  XCTAssert(threw);
+}
+
+- (void)testSchemesNotSupportedException {
+  [_fakeMainBundle fakeMissingAllSchemes];
```
mdmathias, line 92: You may not need this if you do the start faking calls in the test methods where that is needed.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+                                                  completion:nil]);
+}
+
+- (void)testClientIDMissingException {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnonnull"
+  _verifyAccountDetail.configuration = [[GIDConfiguration alloc] initWithClientID:nil];
+#pragma GCC diagnostic pop
+  BOOL threw = NO;
+  @try {
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+                          @"You must specify |clientID| in |GIDConfiguration|");
+  } @finally {
+  }
+  XCTAssert(threw);
+}
+
+- (void)testSchemesNotSupportedException {
+  [_fakeMainBundle fakeMissingAllSchemes];
+  BOOL threw = NO;
+  @try {
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+#pragma GCC diagnostic pop
+  BOOL threw = NO;
+  @try {
+    [_verifyAccountDetail verifyAccountDetails:_verifiableAccountDetails
+                      presentingViewController:_presentingViewController
+                                    completion:nil];
+  } @catch (NSException *exception) {
+    threw = YES;
+    XCTAssertEqualObjects(exception.description,
+                          @"You must specify |clientID| in |GIDConfiguration|");
```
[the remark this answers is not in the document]
brnnmrls: Same as below: Seems that I'm unable to pass in a string variable into the 'reason' parameter in `XCTAssertThrowsSpecificNamed()`. It throws an error and will only succeed when it takes a [literal](https://developer.apple.com/documentation/xctest/xctassertthrows) but maybe I'm misunderstanding.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Added tests for when config is missing, added, and invalid. Let me know if I missed any.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Removed, thanks!

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiableAccountDetail.h"
+
+#import "GoogleSignIn/Tests/Unit/GIDFakeMainBundle.h"
+
+static NSString * const kClientId = @"FakeClientID";
+
+@interface GIDVerifyAccountDetailTests : XCTestCase {
+@private
+  // The |UIViewController| object being tested.
+  UIViewController *_presentingViewController;
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: The variants are calling `- [GIDVerifyAccountDetail init] ` which then calls and is testing for `- [GIDVerifyAccountDetail initWithConfig:]`.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -58,8 +73,58 @@ - (void)verifyAccountDetails:(NSArray<GIDVerifiableAccountDetail *> *)accountDet
...
+  GIDSignInCallbackSchemes *schemes =
+  [[GIDSignInCallbackSchemes alloc] initWithClientIdentifier:options.configuration.clientID];
+  NSArray<NSString *> *unsupportedSchemes = [schemes unsupportedSchemes];
+  if (unsupportedSchemes.count != 0) {
+    // NOLINTNEXTLINE(google-objc-avoid-throwing-exception)
+    [NSException raise:NSInvalidArgumentException
+                format:@"Your app is missing support for the following URL schemes: %@",
+     [unsupportedSchemes componentsJoinedByString:@", "]];
+  }
+  // TODO(#397): Start the incremental authorization flow.
```
mdmathias, line 107: Since there needs to be a current user, perhaps we should check this as well? If no user, then an error needs to be given back to that effect.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,144 @@
...
+  GIDVerifyAccountDetail *verifyAccountDetail = [[GIDVerifyAccountDetail alloc] init];
+  XCTAssertNotNil(verifyAccountDetail.configuration);
+  XCTAssertEqual(verifyAccountDetail.configuration.clientID, kClientId);
+  XCTAssertNil(verifyAccountDetail.configuration.serverClientID);
+  XCTAssertNil(verifyAccountDetail.configuration.hostedDomain);
+  XCTAssertNil(verifyAccountDetail.configuration.openIDRealm);
+}
+
+- (void)testInit_noConfig {
+  [_fakeMainBundle startFakingWithClientID:kClientId];
```
mdmathias, line 75: I think you could pass nil for the client ID here and skip what's below. `GIDFakeMainBundle.h` doesn't have nullability annotations...

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,144 @@
...
+                       hostedDomain:nil
+                        openIDRealm:nil];
+  GIDVerifyAccountDetail *verifyAccountDetail = [[GIDVerifyAccountDetail alloc] init];
+
+  XCTAssertNil(verifyAccountDetail.configuration);
+}
+
+
+- (void)testInit_fullConfig {
+  [_fakeMainBundle startFakingWithClientID:kClientId];
```
mdmathias, line 87: I think you would flip these two calls. Set up the fake with the data, and then tell the fake to start faking. It probably works both ways, but this way feels more logical to me.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,144 @@
...
+  GIDVerifyAccountDetail *verifyAccountDetail = [[GIDVerifyAccountDetail alloc] init];
+  XCTAssertNotNil(verifyAccountDetail.configuration);
+  XCTAssertEqual(verifyAccountDetail.configuration.clientID, kClientId);
+  XCTAssertEqual(verifyAccountDetail.configuration.serverClientID, kServerClientId);
+  XCTAssertEqual(verifyAccountDetail.configuration.hostedDomain, kFakeHostedDomain);
+  XCTAssertEqual(verifyAccountDetail.configuration.openIDRealm, kOpenIDRealm);
+}
+
+- (void)testInit_invalidConfig {
+  [_fakeMainBundle startFakingWithClientID:kClientId];
```
mdmathias, line 102: Same comment about flipping these two calls.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,144 @@
...
+}
+
+- (void)testClientIDMissingException {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnonnull"
+  _verifyAccountDetail.configuration = [[GIDConfiguration alloc] initWithClientID:nil];
+#pragma GCC diagnostic pop
+
+  XCTAssertThrowsSpecificNamed(
+                               [_verifyAccountDetail verifyAccountDetails:_verifiableAccountDetails presentingViewController:_presentingViewController completion:nil],
```
mdmathias, line 127: Formatting: line break and align on the colons.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,144 @@
...
+                               NSException,
+                               NSInvalidArgumentException,
+                               @"You must specify |clientID| in |GIDConfiguration|");
+}
+
+- (void)testSchemesNotSupportedException {
+  XCTAssertThrowsSpecificNamed(
+                               [_verifyAccountDetail verifyAccountDetails:_verifiableAccountDetails presentingViewController:_presentingViewController completion:nil],
```
mdmathias, line 135: Formatting: line break and align on the colons.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
mdmathias: Yes, please add tests where the `GIDConfiguration` is passed in and not created automatically (via a fake).

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,110 @@
...
+  [_fakeMainBundle fakeMissingAllSchemes];
+  BOOL threw = NO;
+  @try {
+    [_verifyAccountDetail verifyAccountDetails:_verifiableAccountDetails
+                      presentingViewController:_presentingViewController
+                                    completion:nil];
+  } @catch (NSException *exception) {
+    threw = YES;
+    XCTAssertEqualObjects(exception.description,
+                          @"Your app is missing support for the following URL schemes: "
```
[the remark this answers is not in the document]
mdmathias: The reason description should be a literal. I'm asking that you make the second comparison string be a constant in this file.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Seems that I get an error when I flip them. This is because `-[GIDFakeMainBundle startFakingWithClientID:]` initializes `NSMutableDictionary *_fakeConfig` and filling the fake with data before initializing will error since the `_fakeConfig` is null. I'll keep it as is unless there's another way to go about it.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Same response as above where errors are thrown if calls are flipped.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Unable to pass nil for the client ID here since an exception will be thrown in [GIDFakeMainBundle](https://github.com/google/GoogleSignIn-iOS/blob/ec20caf39917e3a50905776db1713b37c6b33840/GoogleSignIn/Tests/Unit/GIDFakeMainBundle.m#L49) and . This is because `NSDictionary` does not allow nil values so it throws an exception on initialization. I believe that's why `kClientId` should be passed in to first initialize and then the values in the dictionary can be set to nil.

brnnmrls: An alternative could be to remove line `[_fakeMainBundle startFakingWithClientID:kClientId];` and only keep the second call but I understand this still runs into the issue with nullability annotations. Let me know what you think.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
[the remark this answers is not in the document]
brnnmrls: Gotcha, I think creating the constant is no longer relevant since the assert method has switched to `XCTAssertThrowsSpecificNamed` and wouldn't use the constant, but let me know if I misunderstood.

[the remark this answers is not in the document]
mdmathias: Agreed. Resolving.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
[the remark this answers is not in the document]
mdmathias: Resolving as outdates since we've moved to `XCTAssertThrowsSpecificNamed`.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
mdmathias: You could update `GIDFakeMainBundle` to take a `nullable` string for the `clientId`:

```
- (void)startFakingWithClientID:(nullable NSString *)clientId {
  _clientId = clientId;

  _fakedKeys = @[ kCFBundleURLTypesKey,
                  kConfigClientIDKey,
                  kConfigServerClientIDKey,
                  kConfigHostedDomainKey,
                  kConfigOpenIDRealmKey ];

  _fakeConfig = clientId ? [@{ @"GIDClientID": clientId } mutableCopy] : [@{} mutableCopy];

  [GULSwizzler swizzleClass:[NSBundle class]
                   selector:@selector(objectForInfoDictionaryKey:)
            isClassSelector:NO
                  withBlock:^id(id _self, NSString *key) {
    if ([self->_fakedKeys containsObject:key]) {
      return self->_fakeConfig[key];
    } else {
      @throw [NSException exceptionWithName:@"Requested unexpected info.plist key."
                                     reason:nil
                                   userInfo:nil];
    }
  }];
}
```

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
mdmathias: You could add an initializer that sets up the config. The initializer could take all of the parameters in the call starting on line 88 below. This would set up the config. Then, all that is needed is to call `[_fakeMainBundle startFaking]` (you could add a `-startFaking` method that doesn't take a client ID).

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Added, please let me know if I missed any.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Added check and test.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Good idea, done!

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -43,6 +43,9 @@ typedef void (^GIDVerifyCompletion)(GIDVerifiedAccountDetailResult *_Nullable ve
 /// The active configuration for this instance of `GIDVerifyAccountDetail`.
 @property(nonatomic, nullable) GIDConfiguration *configuration;
 
+- (instancetype)initWithConfig:(nullable GIDConfiguration *)config
```
mdmathias, line 46: Please add doc comments.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -286,12 +286,12 @@ - (void)setUp {
 #elif TARGET_OS_OSX
   _presentingWindow = OCMStrictClassMock([NSWindow class]);
 #endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
-  _authState = OCMStrictClassMock([OIDAuthState class]);
+  _authState = OCMClassMock([OIDAuthState class]);
```
mdmathias, line 289: Please add a link to the issue describing the problem we are addressing here.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,222 @@
...
+static NSString * const kMissingPresentingViewControllerExceptionReason =
+    @"|presentingViewController| must be set.";
+
+static NSString * const kClientId = @"FakeClientID";
+static NSString * const kServerClientId = @"FakeServerClientID";
+static NSString * const kOpenIDRealm = @"FakeRealm";
+static NSString * const kFakeHostedDomain = @"fakehosteddomain.com";
+
+@interface GIDVerifyAccountDetailTests : XCTestCase
+// The |UIViewController| object being tested.
```
mdmathias, line 44: Use `///` here and below.

mdmathias, line 44: Use `///` here and below.

mdmathias, line 44: Use `///` here and below.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,222 @@
...
+  XCTAssertNil(verifyAccountDetail.configuration);
+}
+
+- (void)testInitWithConfig_fullConfig {
+  GIDConfiguration *configuration = [[GIDConfiguration alloc] initWithClientID:kClientId
+                                                                serverClientID:kServerClientId
+                                                                  hostedDomain:kFakeHostedDomain
+                                                                   openIDRealm:kOpenIDRealm];
+
+  GIDVerifyAccountDetail *verifyAccountDetail = [[GIDVerifyAccountDetail alloc]
```
mdmathias, line 132: Wrap to next line after `=`:

```
GIDVerifyAccountDetail *verifyAccountDetail = 
    [[GIDVerifyAccountDetail alloc] initWithConfig:configuration];
```

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
mdmathias, line 46: Also, perhaps I'm forgetting, but should `config` be `nullable`? Isn't the purpose of using this initializer to pass in a `GIDConfiguration`? If passing nil, why not just use the parameter-less initializer? Maybe I'm missing a use-case...

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.h
```diff
@@ -22,12 +22,31 @@
...
 
+/**
+ * @fn initWithClientID:serverClientID:hostedDomain:openIDRealm:
+ * @brief Initializes a GIDFakeMainBundle object.
+ * @param clientID The fake client idenfitier for the app.
+ * @param serverClientID The fake server client idenfitier for the app.
+ * @param hostedDomain The fake hosted domain for the app.
+ * @param openIDRealm The fake OpenID realm for the app.
+ */
+- (nullable instancetype)initWithClientID:(nullable id)clientID
```
mdmathias, line 33: I think we want this to be the designated initializer. Then, we can implement `-init` to call thru to this, which will pass `nil` to all of the params.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
```diff
@@ -0,0 +1,222 @@
...
+  XCTAssertEqual(verifyAccountDetail.configuration.openIDRealm, kOpenIDRealm);
+}
+
+- (void)testCurrentUserException {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnonnull"
+  _verifyAccountDetail.configuration = [[GIDConfiguration alloc] initWithClientID:nil];
+#pragma GCC diagnostic pop
+
+  OIDAuthState *authState = [OIDAuthState testInstance];
```
mdmathias, line 147: Not used here.

mdmathias (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: I believe this was done following [this comment](https://github.com/google/GoogleSignIn-iOS/pull/401#discussion_r1585529200) earlier in the PR but maybe I misunderstood. I do prefer to keep it as non-nullable as well.

brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.h
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Updated!

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Sounds good, I've implemented this suggestion.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Added tests using a passed `GIDConfiguration`.

GoogleSignIn/Tests/Unit/GIDVerifyAccountDetailTest.m
brnnmrls: Resolving this since its no longer issue by calling the created initializer first then calling `startFaking`.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
mdmathias: Ohhh, I think it's because of the parameter-less init:

```
- (instancetype)init {
  GIDConfiguration *configuration;
  NSBundle *bundle = NSBundle.mainBundle;
  if (bundle) {
    configuration = [GIDConfiguration configurationFromBundle:bundle];
  }

  return [self initWithConfig:configuration];
}
```

The above calls `+[GIDConfiguration configurationFromBundle:]`, which [can return nil](https://github.com/google/GoogleSignIn-iOS/blob/briannamorales/vwg-flow/GoogleSignIn/Sources/GIDConfiguration.m#L78). Thus, `[self initWithConfig:configuration]` can pass in a nil `GIDConfiguration`. 

I think `-[GIDVerifyAccountDetail initWithConfig:]` should _not_ be marked `nullable`. If the configuration is being passed in, then why pass in nil?

Furthermore, the parameter-less `init` should probably be annotated as `nullable` and return nil if the automatically created `GIDConfiguration` is nil. We need the configuration.

brnnmrls: Ahh makes sense, added!

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -43,6 +43,13 @@ typedef void (^GIDVerifyCompletion)(GIDVerifiedAccountDetailResult *_Nullable ve
 /// The active configuration for this instance of `GIDVerifyAccountDetail`.
 @property(nonatomic, nullable) GIDConfiguration *configuration;
 
+/// Initialize a `GIDVerifyAccountDetail` object by specifying all available properties.
+///
+/// @param config The configuration to be used.
+/// @return An initialized `GIDVerifyAccountDetail` instance.
+- (instancetype)initWithConfig:(GIDConfiguration *)config
```
mdmathias, line 50: Needs to be `nullable` as well.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -43,6 +43,13 @@ typedef void (^GIDVerifyCompletion)(GIDVerifiedAccountDetailResult *_Nullable ve
 /// The active configuration for this instance of `GIDVerifyAccountDetail`.
 @property(nonatomic, nullable) GIDConfiguration *configuration;
 
+/// Initialize a `GIDVerifyAccountDetail` object by specifying all available properties.
+///
+/// @param config The configuration to be used.
+/// @return An initialized `GIDVerifyAccountDetail` instance.
```
mdmathias, line 49: This should say something about how and why the initializer will return nil.

mdmathias (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Added.

mdmathias: I don't think this initializer should return nil. It's the parameter-less `init` that can return nil if we cannot make a `GIDConfiguration` automatically from the main bundle. 

Sorry for the churn here!

mdmathias: Please update the nullability annotation and documentation here.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,6 +27,30 @@
 
 @implementation GIDVerifyAccountDetail
 
+- (nullable instancetype)initWithConfig:(GIDConfiguration *)configuration {
```
mdmathias, line 30: No need for `nullable`.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -27,6 +27,30 @@
...
+  self = [super init];
+  if (self) {
+    _configuration = configuration;
+  }
+  return self;
+}
+
+- (nullable instancetype)init {
```
mdmathias, line 38: Please add this to the `.h` and document what the initializer does, including how it can return nil.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.m
```diff
@@ -37,16 +37,38 @@ @implementation GIDFakeMainBundle {
   NSMutableDictionary *_fakeConfig;
 }
 
-- (void)startFakingWithClientID:(NSString *)clientId {
-  _clientId = clientId;
+- (nullable instancetype)initWithClientID:(nullable id)clientID
```
mdmathias, line 40: Doesn't need `nullable` here.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.h
```diff
@@ -22,12 +22,32 @@
...
```
mdmathias, line 33: No need for `nullable`.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.m
```diff
@@ -37,16 +37,38 @@ @implementation GIDFakeMainBundle {
...
+
+    [self fakeWithClientID:clientID
+            serverClientID:serverClientID
+              hostedDomain:hostedDomain
+               openIDRealm:openIDRealm];
+  }
+  return self;
+}
+
+- (instancetype)init {
```
mdmathias, line 59: Please add to the `.h` and add docs that we pass nil to all of the designated initializer's parameters.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDFakeMainBundle.h
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Done.

brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -43,6 +43,21 @@ typedef void (^GIDVerifyCompletion)(GIDVerifiedAccountDetailResult *_Nullable ve
...
+- (instancetype)initWithConfig:(GIDConfiguration *)config
+    NS_DESIGNATED_INITIALIZER;
+
+
+/// Initialize a `GIDVerifyAccountDetail` object by calling the designated initializer 
+/// with the default configuration from the bundle's Info.plist.
+///
+/// @return An initialized `GIDVerifyAccountDetail` instance.
+///     Otherwise, `nil` if the configuration is invalid .
```
mdmathias, line 58: nit: remove space before period.

Also suggest replacing "Otherwise, `nil` if the configuration is invalid ." with "Otherwise, `nil` if the configuration cannot be automatically generated from your app's Info.plist."

mdmathias (approved):
Just a small comment to improve a doc comment.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Updated!

PR #403, opened 2024-04-30
Remove GTMSessionFetcher module import

### LINKED ISSUE #396: Upgrading to 7.1.0 from 7.0.0, now we can't `po` anything in lldb due to GTMSessionFetcher import issue.
**Describe the bug**

We are seeing issues in `lldb` related to `GTMSessionFetcher` since updating to 7.1.0, where it complains it cannot import GTMSessionFetcher any time we try to `po` a value from our code in the debugger — which is obviously very bad news.

I'm far from an expert on this but in our case the problem "goes away" if we change the conditional import in GIDGoogleUser.h https://github.com/google/GoogleSignIn-iOS/blob/ec20caf39917e3a50905776db1713b37c6b33840/GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h#L26 so that it *never* uses the `@import` modular import variant.

Why this is suddenly a problem - as this has been in the GoogleSignIn code for a while – is not clear. We relatively recently (months back) added one SPM package to our app. The rest are all Cocoapods, including GSI and its dependencies. Perhaps this has triggered the problem, but we certainly only noticed it recently when switching from 7.0.0 to 7.1.0.

The clear implication is that `SWIFT_PACKAGE` is now defined for our builds where it wasn't previously, but this is breaking lldb for some reason, perhaps because GSI is still built with Cocoapods, not SPM?

**To Reproduce**

Sorry, we don't know.

**Expected behavior**

We should not have any issues using the debugger, least of all complaints about failure of lldb to import GTMSessionFetcher.

**Screenshots**

**Environment**
- Browser: Xcode 15.2 / Xcode 15.3

**Additional context**

mdmathias:
This change removes the modular import of `GTMSessionFetcher` in `GIDGoogleUser.h` (e.g., `@import GTMSessionFetcher`) when building under `SWIFT_PACKAGE` in favor of `#import <GTMSessionFetcher/GTMSessionFetcher.h>`. The intention is to address lldb crashes when `po`ing in the debug console when projects use both SPM and CocoaPods.

Fixes #396

### DIFF
--- a/GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
+++ b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h
@@ -23,11 +23,7 @@
 #import <AppKit/AppKit.h>
 #endif
 
-#ifdef SWIFT_PACKAGE
-@import GTMSessionFetcherCore;
-#else
 #import <GTMSessionFetcher/GTMSessionFetcher.h>
-#endif
 
 @class GIDConfiguration;
 @class GIDSignInResult;


mdmathias:
Here is a toy sample project that uses both SPM and CocoaPods where GSI is a dep in CocoaPods using this branch. I am able to `po` without crashing LLDB.

[LLDBCrash.zip](https://github.com/google/GoogleSignIn-iOS/files/15169595/LLDBCrash.zip)

yalkhder (approved):

PR #404, opened 2024-05-02
Update SignInView.swift

movsianikov:
Fix issue with declaration of GoogleSignInButtonViewModel variable. Switching to @StateObject from @ObservedObject allows to customize the button using constructor, otherwise - the customization doesn't work.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Views/SignInView.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Views/SignInView.swift
@@ -19,7 +19,7 @@ import GoogleSignInSwift
 
 struct SignInView: View {
   @EnvironmentObject var authViewModel: AuthenticationViewModel
-  @ObservedObject var vm = GoogleSignInButtonViewModel()
+  @StateObject var vm = GoogleSignInButtonViewModel()
 
   var body: some View {
     VStack {


mdmathias:
Can you clarify what you mean by "allows to customize the button using constructor, otherwise - the customization doesn't work?" I am not sure what you're fixing.

movsianikov:
> Can you clarify what you mean by "allows to customize the button using constructor, otherwise - the customization doesn't work?" I am not sure what you're fixing.

Hi Matt, of course! Here you go:
1. With the current approach this customization does **NOT** work:
`@ObservedObject private var vm = GoogleSignInButtonViewModel(scheme: .dark, style: .wide, state: .normal)`
2. In order to make it work (workaround), I've customized it in the init() method like this:
```
    init() {
        vm.style = .wide
        vm.scheme = .dark
    }
```
3. But the correct way, since the instance of the GoogleSignInButtonViewModel created in the SignInView, would be to declare it as @StateObject, then the customization work fine without init() method:
`@ObservedObject private var vm = GoogleSignInButtonViewModel(scheme: .dark, style: .wide, state: .normal)`

Please let me know if you have any questions, comments, or concerns.

Thank you!

Best regards,
**MAKSYM OVSIANIKOV**

mdmathias:
@movsianikov I'm sorry! I still don't understand what you mean by the "customization" not working. In `SignInView.swift`, if you create the `vm` with a different scheme (e.g., `.dark`), then the button will be drawn with the dark blue background.

```
struct SignInView: View {
  @EnvironmentObject var authViewModel: AuthenticationViewModel
  @ObservedObject var vm = GoogleSignInButtonViewModel(scheme: .dark)
...
```

Screenshot 2024-05-03 at 9 17 49 AM

mdmathias:
I am closing this pull request due to inactivity.

PR #414, opened 2024-05-13
Process authorization response and token fetch.

brnnmrls:
Once an instance of `OIDAuthorizationRequest` is made to start the incremental authorization flow, next step is to process the response made from this request. This is done similar to [the interactive sign in flow](https://github.com/google/GoogleSignIn-iOS/blob/4b4f14b774080066c7351d636e7e8be0441bed55/GoogleSignIn/Sources/GIDSignIn.m#L618). VwG adds an instance method `-[GIDVerifyAccountDetail processAuthorizationResponse:error:]` checking for whether or not the authorization response has an authorization code. Testing will be done in a future PR once the response has been handled.


GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -32,6 +32,19 @@ NS_ASSUME_NONNULL_BEGIN
 @class GIDVerifiableAccountDetail;
 @class GIDVerifiedAccountDetailResult;
 
+/// The error domain for `NSError`s returned by the Google Sign-In SDK.
+extern NSErrorDomain const kGIDVerifyErrorDomain;
+
+/// A list of potential error codes returned from the Google Sign-In SDK.
+typedef NS_ERROR_ENUM(kGIDVerifyErrorDomain, GIDVerifyErrorCode) {
+  /// Indicates an unknown error has occurred.
+  kGIDVerifyErrorCodeUnknown = -1,
```
mdmathias, line 41: Let's make these positive integers.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -32,6 +32,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+  /// Indicates the user canceled the verification request.
+  kGIDVerifyErrorCodeCanceled = -2,
+  /// Indicates an Enterprise Mobility Management related error has occurred.
+  kGIDVerifyErrorCodeEMM = -3,
```
mdmathias, line 45: I don't think EMM is involved...

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -47,9 +61,36 @@
 // Expected path in the URL scheme to be handled.
 static NSString *const kBrowserCallbackPath = @"/oauth2callback";
 
+// The error code for Google Identity.
+NSErrorDomain const kGIDVerifyErrorDomain = @"com.google.GIDVerify";
```
mdmathias, line 65: What do you think about keeping this as `com.google.GIDSignIn`? (FWIW You can read more about [error domains here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html#//apple_ref/doc/uid/TP40001806-CH202-CJBGAIBJ).)

If we will keep the existing domain (which should get moved to the constants file from #401), then `com.google.GIDSignIn` is the domain for all errors from this library. That means the specific error codes will identify the error. One weakness I see with this approach is that the error codes are just integer backed enum cases, so we could have some collision across errors if two error enum cases are backed by the same integer.

What do you think?

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -47,9 +61,36 @@
...
+
+// Error string for user cancelations.
+static NSString *const kUserCanceledError = @"The user canceled the verification flow.";
+
+// Parameters in the callback URL coming back from browser.
+static NSString *const kOAuth2ErrorKeyName = @"error";
+static NSString *const kOAuth2AccessDenied = @"access_denied";
+
+// Minimum time to expiration for a restored access token.
+static const NSTimeInterval kMinimumRestoredAccessTokenTimeToExpire = 600.0;
```
mdmathias, line 75: Just double checking that this is the correct time to expire.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -47,9 +61,36 @@
...
+
+// The callback queue used for authentication flow.
+@interface GIDVerifyAuthFlow : GIDCallbackQueue
```
mdmathias, line 78: I think I would prefer that this lives in its own set of files. Also, remember to replace `//` with `///` if you want a doc comment.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -58,9 +99,9 @@ - (nullable instancetype)initWithConfig:(GIDConfiguration *)configuration {
     _configuration = configuration;
 
     NSString *authorizationEndpointURL = [NSString stringWithFormat:kAuthorizationURLTemplate,
-        [GIDSignInPreferences googleAuthorizationServer]];
+                                          [GIDSignInPreferences googleAuthorizationServer]];
```
mdmathias, line 102: I think everything can fit on one line below the `=`.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -58,9 +99,9 @@ - (nullable instancetype)initWithConfig:(GIDConfiguration *)configuration {
...
     NSString *tokenEndpointURL = [NSString stringWithFormat:kTokenURLTemplate,
-        [GIDSignInPreferences googleTokenServer]];
+                                  [GIDSignInPreferences googleTokenServer]];
```
mdmathias, line 104: Same comment about everything fitting on one line.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
```diff
@@ -168,18 +210,115 @@ - (void)verifyAccountDetailsInteractivelyWithOptions:(GIDSignInInternalOptions *
...
+      errorString = kUserCanceledError;
+      errorCode = kGIDVerifyErrorCodeCanceled;
+    }
+    authFlow.error = [self errorWithString:errorString code:errorCode];
+  }
+
+  // TODO: Add completion callback method (#413).
+}
+
+// Fetches the access token if necessary as part of the auth flow.
```
mdmathias, line 267: I don't think this is necessary.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: I also thought of keeping it as `com.google.GIDSignIn` since I liked the idea of the domain covering the library as a whole but felt being more specific on the error domain wouldn't hurt. Though, I feel that keeping the domain as is would work best if we're able to keep the integer ranges of both enum cases separate. For example, the verify error codes could start at 1000.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Oops missed this one, my bad, removed!

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
mdmathias: Seems like keeping this at 600.0 is okay. Resolving.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifyAccountDetail.m
mdmathias: We're in a kind of tricky spot since `GIDSignInErrorCode`s are negative. :/ 

I think maybe it's better to (for now) to go with `com.google.GIDVerifyAccountDetail` for the domain. It's cleaner that way, IMHO.

If you agree, would you mind creating an issue to revisit and finalize our position on this and link here in the code?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
```diff
@@ -32,6 +32,17 @@ NS_ASSUME_NONNULL_BEGIN
 @class GIDVerifiableAccountDetail;
 @class GIDVerifiedAccountDetailResult;
 
+/// The error domain for `NSError`s returned by the Google Sign-In SDK.
+extern NSErrorDomain const kGIDVerifyErrorDomain;
+
+/// A list of potential error codes returned from the Google Sign-In SDK.
+typedef NS_ERROR_ENUM(kGIDVerifyErrorDomain, GIDVerifyErrorCode) {
+  /// Indicates an unknown error has occurred.
+  kGIDVerifyErrorCodeUnknown = 1,
```
mdmathias, line 41: I think I prefer not having the `k` prefix in the case names. Those aren't in the [style guide](https://google.github.io/styleguide/objcguide.html#constants).

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
```diff
@@ -0,0 +1,76 @@
...
+@interface GIDAuthorizationResponseHelper : NSObject
+
+/// The authorization response to process.
+@property(nonatomic, readonly) OIDAuthorizationResponse *authorizationResponse;
+
+/// The EMM support version.
+@property(nonatomic, readwrite, nullable) NSString *emmSupport;
+
+/// The name of the current flow.
+@property(nonatomic, readonly)GIDFlowName flowName;
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
```diff
@@ -0,0 +1,76 @@
...
+#import <Foundation/Foundation.h>
+
+@class GIDAuthFlow;
+@class GIDConfiguration;
+@class OIDAuthorizationResponse;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A list of potential current flow names.
+typedef NS_ENUM(NSInteger, GIDFlowName) {
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
```diff
@@ -0,0 +1,76 @@
...
+/// The authorization response to process.
+@property(nonatomic, readonly) OIDAuthorizationResponse *authorizationResponse;
+
+/// The EMM support version.
+@property(nonatomic, readwrite, nullable) NSString *emmSupport;
+
+/// The name of the current flow.
+@property(nonatomic, readonly)GIDFlowName flowName;
+
+/// The configuration of the current flow.
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthFlow.h
```diff
@@ -0,0 +1,41 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "GoogleSignIn/Sources/GIDCallbackQueue.h"
+
+@class OIDAuthState;
+@class GIDProfileData;
+
+/// The callback queue used for authentication flow.
+@interface GIDAuthFlow : GIDCallbackQueue
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
```diff
@@ -0,0 +1,76 @@
...
+
+/// Initializes a new instance of the `GIDAuthorizationResponseHelper` class with the provided fields.
+///
+/// @param authorizationResponse The authorization response to be processed.
+/// @param emmSupport The EMM support version.
+/// @param flowName The name of the current flow.
+/// @param configuration The configuration.
+/// @return A new initialized instance of the `GIDAuthorizationResponseHelper` class.
+- (instancetype)
+    initWithAuthorizationResponse:(nullable OIDAuthorizationResponse *)authorizationResponse
```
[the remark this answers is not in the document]
brnnmrls: Makes sense, added! In regard to simplifying the impl code, do you mind clarifying your vision on what could change?

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
```diff
@@ -0,0 +1,76 @@
...
+/// @param error The error thrown if there's no authorization response.
+/// @return An instance of `GIDAuthFlow`.
+- (GIDAuthFlow *)processWithError:(NSError *)error;
+
+
+
+/// Fetches the access token if necessary as part of the auth flow.
+///
+/// @param authFlow The auth flow to either fetch tokens or error.
+- (void)maybeFetchToken:(GIDAuthFlow *)authFlow;
```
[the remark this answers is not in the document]
brnnmrls: Yep, exactly. I'm calling this from `GIDSignIn`'s non-interactive flow.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+#endif
+
+/// Error string for user cancelations.
+static NSString *const kUserCanceledSignInError = @"The user canceled the sign-in flow.";
+static NSString *const kUserCanceledVerifyError = @"The user canceled the verification flow.";
+
+/// Minimum time to expiration for a restored access token.
+static const NSTimeInterval kMinimumRestoredAccessTokenTimeToExpire = 600.0;
+
+@implementation GIDAuthorizationResponseHelper
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+    } else {
+      [authFlow next];
+    }
+#elif TARGET_OS_OSX || TARGET_OS_MACCATALYST
+    [authFlow next];
+#endif // TARGET_OS_OSX || TARGET_OS_MACCATALYST
+  }];
+}
+
+- (void)authorizationCodeErrorWithAuthFlow:(GIDAuthFlow *)authFlow {
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+      }
+
+      authFlow.error = [self errorWithString:errorString code:errorCode];
+#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+      break;
+    }
+  }
+}
+
+- (void)authorizationResponseErrorWithAuthFlow:(GIDAuthFlow *)authFlow
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,52 @@
...
+
+#import "GoogleSignIn/Tests/Unit/OIDAuthorizationResponse+Testing.h"
+
+@interface GIDAuthorizationResponseHelperTest : XCTestCase
+@end
+
+@implementation GIDAuthorizationResponseHelperTest
+
+- (void)testInitWithAuthorizationResponse {
+  // Mock generating a GIDConfiguration when initializing GIDGoogleUser.
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h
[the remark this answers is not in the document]
mdmathias: I believe you check for whether or not there is an `_authorizationResponse` somewhere in the impl. You won't need that if it's non-null.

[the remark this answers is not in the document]
mdmathias: Actually ignore my suggestion to not check: `if (_authorizationResponse) { ...`. We should still have the param marked `nonnull`, but the check in the impl is needed since the nullability annotation does not guarantee that the passed argument is `nonnull`.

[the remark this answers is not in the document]
brnnmrls: Sounds good! Thanks for double checking.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+      authState.lastAuthorizationResponse.authorizationCode) {
+    tokenRequest = [authState.lastAuthorizationResponse
+        tokenExchangeRequestWithAdditionalParameters:additionalParameters];
+  } else {
+    [additionalParameters
+        addEntriesFromDictionary:authState.lastTokenResponse.request.additionalParameters];
+    tokenRequest = [authState tokenRefreshRequestWithAdditionalParameters:additionalParameters];
+  }
+
+  [authFlow wait];
```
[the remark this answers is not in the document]
brnnmrls: As discussed offline, I didn't run into issues running through the manual testing plan and can confirm that the sign-in flow works with the new implementation. I went ahead and added the newly created issue as a `TODO`.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+      [self authorizationCodeErrorWithAuthFlow:authFlow];
+    }
+  } else {
+    [self authorizationResponseErrorWithAuthFlow:authFlow
+                                           error:error];
+  }
+  return authFlow;
+}
+
+- (void)maybeFetchToken:(GIDAuthFlow *)authFlow {
```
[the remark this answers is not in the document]
brnnmrls: Same as comment below where we will clean up the callback flow in the future.

GoogleSignIn/Sources/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,253 @@
...
+  if (self) {
+    _authorizationResponse = authorizationResponse;
+    _emmSupport = emmSupport;
+    _flowName = flowName;
+    _configuration = configuration;
+  }
+  return self;
+}
+
+- (GIDAuthFlow *)processWithError:(NSError *)error {
```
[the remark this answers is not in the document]
brnnmrls: Is this comment still relevant? I imagine it's probably a good idea to change the name regardless.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,52 @@
...
+#import "GoogleSignIn/Sources/GIDAuthorizationResponseHelper.h"
+
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDConfiguration.h"
+
+#import "GoogleSignIn/Tests/Unit/OIDAuthorizationResponse+Testing.h"
+
+@interface GIDAuthorizationResponseHelperTest : XCTestCase
+@end
+
+@implementation GIDAuthorizationResponseHelperTest
```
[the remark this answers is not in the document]
brnnmrls: Acknowledged, currently working on this.

GoogleSignIn/Tests/Unit/GIDAuthFlowTest.m
```diff
@@ -0,0 +1,54 @@
...
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GIDProfileData *profileData = [GIDProfileData testInstance];
+
+  GIDAuthFlow *authFlow = [[GIDAuthFlow alloc] initWithAuthState:authState
+                                                           error:nil
+                                                      emmSupport:nil
+                                                     profileData:profileData];
+
+  XCTAssertNotNil(authFlow);
+  XCTAssertNotNil(authFlow.authState);
```
mdmathias, line 38: I suggest that you assert equality instead of not nil.

GoogleSignIn/Tests/Unit/GIDAuthFlowTest.m
```diff
@@ -0,0 +1,54 @@
...
+  XCTAssertNil(authFlow.error);
+  XCTAssertNil(authFlow.emmSupport);
+  XCTAssertNotNil(authFlow.profileData);
```
mdmathias, line 41: Same: assert equality.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  OIDAuthState *authState = [OIDAuthState testInstance];
+  GIDAuthorizationResponseHandlingFake *fakeHandler = 
+      [[GIDAuthorizationResponseHandlingFake alloc] initWithAuthState:authState error:nil];
+  GIDAuthorizationResponseHelper *responseHelper =
+      [[GIDAuthorizationResponseHelper alloc] initWithAuthorizationResponseHandler:fakeHandler];
+  GIDAuthFlow *authFlow = [[GIDAuthFlow alloc] init];
+
+  [responseHelper fetchTokenWithAuthFlow:authFlow];
+  XCTAssertNotNil(authFlow);
+  XCTAssertNotNil(authFlow.authState);
```
mdmathias, line 62: Assert equality. Not nil is useful, but it's more important to know that the `authState` on line 53 is the same as `authState.authState` here.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  GIDAuthorizationResponseHandler *responseHandler =
+      [[GIDAuthorizationResponseHandler alloc] initWithAuthorizationResponse:authorizationResponse
+                                                                  emmSupport:nil
+                                                                    flowName:GIDFlowNameVerify
+                                                               configuration:nil
+                                                                       error:nil];
+
+  GIDAuthFlow *authFlow = [responseHandler generateAuthFlowFromAuthorizationResponse];
+  XCTAssertNotNil(authFlow);
+  XCTAssertNotNil(authFlow.authState);
```
mdmathias, line 77: Can we verify that this `authState` is what we expect it to be?

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  XCTAssertNil(authFlow.error);
+}
+
+- (void)testGenerateAuthFlowFromAuthorizationResponse_noCode {
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
```
mdmathias, line 82: I think this check should be outside of the test so that the test doesn't even run outside of these listed targets.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  XCTAssertNil(authFlow.authState);
+  XCTAssertNotNil(authFlow.error);
+  XCTAssertEqual(authFlow.error.code, expectedError.code);
+  XCTAssertEqualObjects(authFlow.error.userInfo[NSLocalizedDescriptionKey],
+                        expectedError.userInfo[NSLocalizedDescriptionKey]);
+#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+}
+
+- (void)testGenerateAuthFlowWithMissingAuthorizationResponse {
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
```
mdmathias, line 110: Move outside.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+                                                                  emmSupport:nil
+                                                                    flowName:GIDFlowNameVerify
+                                                               configuration:nil
+                                                                       error:error];
+
+  GIDAuthFlow *authFlow = [responseHandler generateAuthFlowFromAuthorizationResponse];
+  XCTAssertNotNil(authFlow);
+  XCTAssertNil(authFlow.authState);
+  XCTAssertNotNil(authFlow.error);
+  XCTAssertEqual(authFlow.error.code, error.code);
```
mdmathias, line 125: I think maybe it would be better to assert that the two errors are the same.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  GIDAuthorizationResponseHandler *responseHandler = 
+      [[GIDAuthorizationResponseHandler alloc] initWithAuthorizationResponse:nil
+                                                                  emmSupport:kEMMVersion
+                                                                    flowName:GIDFlowNameSignIn
+                                                               configuration:nil
+                                                                       error:nil];
+
+  [responseHandler maybeFetchToken:authFlow];
+  XCTAssertNil(authFlow.authState);
+  XCTAssertNotNil(authFlow.error);
```
mdmathias, line 146: Assert that the error is what we expect it to be.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+  XCTAssertNil(authFlow.emmSupport);
+}
+
+- (void)testMaybeFetchTokenWithExpirationError {
```
mdmathias, line 150: I don't understand the name of this test method and what's being done inside of it. I expect some kind of assert around an expiration error, but I don't see that.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,169 @@
...
+                                                     profileData:nil];
+  GIDAuthorizationResponseHandler *responseHandler =
+      [[GIDAuthorizationResponseHandler alloc] initWithAuthorizationResponse:nil
+                                                                  emmSupport:kEMMVersion
+                                                                    flowName:GIDFlowNameSignIn
+                                                               configuration:nil
+                                                                       error:nil];
+
+  [responseHandler maybeFetchToken:authFlow];
+  XCTAssertNotNil(authFlow.authState);
```
mdmathias, line 164: If we do expect an `authState`, then let's assert that it has what we expect on it.

GoogleSignIn/Sources/GIDAuthorizationResponse/API/GIDAuthorizationResponseHandling.h
```diff
@@ -0,0 +1,62 @@
...
+@class OIDAuthorizationResponse;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A list of potential current flow names.
+typedef NS_ENUM(NSInteger, GIDFlowName) {
+/// The Sign In flow.
+  GIDFlowNameSignIn = 0,
+/// The Verify flow.
+  GIDFlowNameVerify = 1,
```
mdmathias, line 30: Can we rename this to `GIDFlowNameVerifyAccountDetail`?

mdmathias (changes requested):

GoogleSignIn/Sources/GIDAuthorizationResponse/GIDAuthorizationResponseHelper.m
```diff
@@ -0,0 +1,69 @@
...
+#import "GoogleSignIn/Sources/GIDEMMErrorHandler.h"
+#import "GoogleSignIn/Sources/GIDEMMSupport.h"
+#import "GoogleSignIn/Sources/GIDSignInConstants.h"
+#import "GoogleSignIn/Sources/GIDSignInPreferences.h"
+
+@import GTMAppAuth;
+
+#ifdef SWIFT_PACKAGE
+@import AppAuth;
+@import GTMSessionFetcherCore;
```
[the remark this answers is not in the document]
brnnmrls: I don't think so, removed.

GoogleSignIn/Sources/GIDAuthorizationResponse/Implementations/GIDAuthorizationResponseHandler.m
```diff
@@ -0,0 +1,273 @@
...
+#import "GoogleSignIn/Sources/GIDEMMErrorHandler.h"
+#import "GoogleSignIn/Sources/GIDEMMSupport.h"
+
+#import "GoogleSignIn/Sources/GIDSignInPreferences.h"
+
+@import GTMAppAuth;
+
+#ifdef SWIFT_PACKAGE
+@import AppAuth;
+@import GTMSessionFetcherCore;
```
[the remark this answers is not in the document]
brnnmrls: Removed.

GoogleSignIn/Tests/Unit/GIDAuthFlowTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthFlowTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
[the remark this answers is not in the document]
brnnmrls: Added.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Gotcha, done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Sounds good, moved.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponse/Fake/GIDAuthorizationResponseHandlingFake.h
```diff
@@ -0,0 +1,43 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "GoogleSignIn/Sources/GIDAuthorizationResponse/API/GIDAuthorizationResponseHandling.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class OIDAuthState;
+
+@interface GIDAuthorizationResponseHandlingFake : NSObject <GIDAuthorizationResponseHandling>
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponse/Implementations/GIDAuthorizationResponseHandler.h
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+#import "GoogleSignIn/Sources/GIDAuthorizationResponse/API/GIDAuthorizationResponseHandling.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GIDAuthorizationResponseHandler : NSObject <GIDAuthorizationResponseHandling>
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Changed to `testMaybeFetchToken_noRefresh` indicating that this will test `-[GIDAuthorizationResponseHandler maybeFetchToken:]` when the access token isn't near expiration. Let me know if you have a suggestion.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponse/API/GIDAuthorizationResponseHandling.h
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -92,78 +101,89 @@ - (void)testGenerateAuthFlowFromAuthorizationResponse_noCode {
...
 
   GIDAuthFlow *authFlow = [responseHandler generateAuthFlowFromAuthorizationResponse];
   XCTAssertNotNil(authFlow);
   XCTAssertNil(authFlow.authState);
   XCTAssertNotNil(authFlow.error);
-  XCTAssertEqual(authFlow.error.code, error.code);
-#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+  XCTAssertEqual(authFlow.error.domain, expectedError.domain);
+  XCTAssertEqual(authFlow.error.code, expectedError.code);
+  XCTAssertEqualObjects(authFlow.error.userInfo, expectedError.userInfo);
```
mdmathias, line 143: Are you able to simplify these to just `XCTAssertEqual(authFlow.error, expectedError)`?

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -92,78 +101,89 @@ - (void)testGenerateAuthFlowFromAuthorizationResponse_noCode {
...
-                                                                  emmSupport:kEMMVersion
-                                                                    flowName:GIDFlowNameSignIn
-                                                               configuration:nil
-                                                                       error:nil];
 
   [responseHandler maybeFetchToken:authFlow];
   XCTAssertNil(authFlow.authState);
   XCTAssertNotNil(authFlow.error);
-  XCTAssertNil(authFlow.emmSupport);
+  XCTAssertEqual(authFlow.error, error);
```
mdmathias, line 166: I think you want to use `XCTAssertEqualObjects`. [XCTAssertEqual](https://developer.apple.com/documentation/xctest/xctassertequal) is for C scalars.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -92,78 +101,89 @@ - (void)testGenerateAuthFlowFromAuthorizationResponse_noCode {
...
-                                                                  emmSupport:kEMMVersion
-                                                                    flowName:GIDFlowNameSignIn
-                                                               configuration:nil
-                                                                       error:nil];
 
   [responseHandler maybeFetchToken:authFlow];
   XCTAssertNotNil(authFlow.authState);
-  XCTAssertNil(authFlow.error);
-  XCTAssertNil(authFlow.emmSupport);
+  XCTAssertEqual(authFlow.authState, authState);
```
mdmathias, line 184: Use `XCTAssertEqualObjects`.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -92,78 +101,89 @@ - (void)testGenerateAuthFlowFromAuthorizationResponse_noCode {
...
+  XCTAssertNotNil(authFlow.error);
+  XCTAssertEqual(authFlow.error, error);
```
mdmathias, line 186: Use `XCTAssertEqualObjects`.

GoogleSignIn/Tests/Unit/OIDTokenResponse+Testing.h
```diff
@@ -53,6 +53,8 @@ extern NSString * const kFatPictureURL;
 
 + (instancetype)testInstanceWithIDToken:(NSString *)idToken;
 
++ (instancetype)testInstanceWithAccessTokenExpiresIn:(NSNumber *)accessTokenExpiresIn;
```
mdmathias, line 56: For your consideration: `testInstanceWithAccessTokenExpiration:`.

Also, should the parameter's type be `NSNumber`? I think you're passing in a `NSTimeInterval`?

mdmathias: Also, you can rename the parameter to `expiration`.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDAuthorizationResponse/API/GIDAuthorizationResponseHandling.h
```diff
@@ -27,7 +27,7 @@ typedef NS_ENUM(NSInteger, GIDFlowName) {
 /// The Sign In flow.
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponse/Implementations/GIDAuthorizationResponseHandler.h
```diff
@@ -20,6 +20,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/// A class that handles the authorization response and fetch tokens.
```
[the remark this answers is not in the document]
brnnmrls: Yes o.o, fixed!

GoogleSignIn/Sources/GIDAuthorizationResponse/Implementations/GIDAuthorizationResponseHandler.m
```diff
@@ -97,7 +93,7 @@ - (GIDAuthFlow *)generateAuthFlowFromAuthorizationResponse {
     }
   } else {
     [self authorizationResponseErrorToAuthFlow:authFlow
-                                           error:_error];
+                                         error:_error];
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDAuthorizationResponse/Implementations/GIDAuthorizationResponseHandler.m
```diff
@@ -179,10 +175,10 @@ - (void)authorizationCodeErrorToAuthFlow:(GIDAuthFlow *)authFlow {
       if (authFlow.emmSupport) {
         [authFlow wait];
         BOOL isEMMError = [[GIDEMMErrorHandler sharedInstance]
-                           handleErrorFromResponse:params
-                           completion:^{
-          [authFlow next];
-        }];
+                           handleErrorFromResponse:params
```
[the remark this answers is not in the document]
brnnmrls: Done.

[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -26,10 +26,14 @@
 
 #import "GoogleSignIn/Tests/Unit/OIDAuthorizationResponse+Testing.h"
 #import "GoogleSignIn/Tests/Unit/OIDAuthState+Testing.h"
+#import "GoogleSignIn/Tests/Unit/OIDTokenResponse+Testing.h"
 
 // The EMM support version
 static NSString *const kEMMVersion = @"1";
 
+// Time interval to use for an expiring access token.
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -26,10 +26,14 @@
...
+static NSTimeInterval kExpiringAccessToken = 20;
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Unfortunately doing this results in an error even with the values being the same. I could create an `isEqual:` if preferred.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Done.

GoogleSignIn/Tests/Unit/OIDTokenResponse+Testing.h
brnnmrls: I think it's okay as `NSNumber` since I'm passing in the variable using the `@()` literal on the `NSTimeInterval` making it the expected type. I followed the way `GIDGoogleUserTest` went about passing in the expiration [here](https://github.com/google/GoogleSignIn-iOS/blob/b0a724e1a8e12cfbe027d8f844a1e6f80f17532f/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m#L190-L193) to keep it consistent. Let me know if I should change.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
```diff
@@ -0,0 +1,185 @@
...
+                                                     profileData:nil];
+
+  [responseHandler maybeFetchToken:authFlow];
+  XCTAssertNil(authFlow.authState);
+  XCTAssertNotNil(authFlow.error);
+  XCTAssertEqualObjects(authFlow.error, error);
+}
+
+- (void)testMaybeFetchToken_noRefresh {
+  NSError *error = [NSError errorWithDomain:kGIDSignInErrorDomain
```
mdmathias, line 166: FWIW I like naming things like `error` here as `expectedError` so it's clear that this is the error we are making sure we have.

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
mdmathias: LGTM

GoogleSignIn/Tests/Unit/OIDTokenResponse+Testing.h
mdmathias: I see. It looks like the implementation calls through to [`+[OIDTokenResponse testInstanceWithIDToken:accessToken:expiresIn:refreshToken:tokenRequest:]`](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Tests/Unit/OIDTokenResponse%2BTesting.m#L68), which expects an `NSNumber`. That's because it gets put into a dictionary.

mdmathias (approved):

GoogleSignIn/Tests/Unit/GIDAuthorizationResponseHelperTest.m
brnnmrls: Sounds good, applied.

PR #445, opened 2024-06-17
Update minimum iOS support to 12

### LINKED ISSUE #443: Bump minimum iOS version to 12
# Context 

Apple requires that a minimum of Xcode 15 is used for App Store submissions. Xcode 15's minimum supported iOS version is 12. Currently, GSI supports iOS 11. 

# To Do 

Given the requirement to submit apps built with Xcode 15, we should increment GSI's baseline iOS support to 12.

# NB

This will be a breaking change since clients will need to update their `Package.swift` and `Podfile` files if they previously supported iOS 11. Therefore, this needs to be included in a version 8.0.0 release of this library.

mdmathias:
Fixes #443


brnnmrls (approved):

Iambarry1:
> Fixes #443

PR #452, opened 2024-07-08
[VwG] Add verification flow example to DaysUntilBirthday Sample App.

brnnmrls:
The following changes add the VwG flow to the DaysUntilBirthday Sample App. Once the user clicks 'Verify My Age' on the UserProfileView and navigates to VerificationView. The verify flow commences and either returns a token response or an error. If there's a returned token response, then the VerificationView prints that it's a verified account along with details of the refresh and access tokens. If an error is returned, then the VerificationView displays "Unverified Account" along with a loading icon.

The refresh feature and exercising the tokens will be sent out in separate PRs.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
+++ b/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
@@ -10,6 +10,8 @@
 		4C6A2BA2321434ACBD2AF201 /* Pods_DaysUntilBirthdayForPod__macOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 926A15D393D684C68E09FB0F /* Pods_DaysUntilBirthdayForPod__macOS_.framework */; };
 		641495072C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
 		641495082C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
+		6414950A2C3E094B00C9A613 /* VerifiedAgeViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495092C3E094B00C9A613 /* VerifiedAgeViewModel.swift */; };
+		6414950C2C3E0C7A00C9A613 /* VerificationLoader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6414950B2C3E0C7A00C9A613 /* VerificationLoader.swift */; };
 		7345AD032703D9470020AFB1 /* DaysUntilBirthday.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7345AD022703D9470020AFB1 /* DaysUntilBirthday.swift */; };
 		7345AD052703D9470020AFB1 /* ContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7345AD042703D9470020AFB1 /* ContentView.swift */; };
 		7345AD072703D9480020AFB1 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 7345AD062703D9480020AFB1 /* Assets.xcassets */; };
@@ -48,6 +50,8 @@
 		29BEB027A694593FA0450863 /* Pods-DaysUntilBirthdayForPod(iOS).debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-DaysUntilBirthdayForPod(iOS).debug.xcconfig"; path = "Target Support Files/Pods-DaysUntilBirthdayForPod(iOS)/Pods-DaysUntilBirthdayForPod(iOS).debug.xcconfig"; sourceTree = "<group>"; };
 		5CF615341A61D92D89389D44 /* Pods-DaysUntilBirthdayForPod (macOS).release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-DaysUntilBirthdayForPod (macOS).release.xcconfig"; path = "Target Support Files/Pods-DaysUntilBirthdayForPod (macOS)/Pods-DaysUntilBirthdayForPod (macOS).release.xcconfig"; sourceTree = "<group>"; };
 		641495062C3C90E100C9A613 /* VerificationView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VerificationView.swift; sourceTree = "<group>"; };
+		641495092C3E094B00C9A613 /* VerifiedAgeViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VerifiedAgeViewModel.swift; sourceTree = "<group>"; };
+		6414950B2C3E0C7A00C9A613 /* VerificationLoader.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VerificationLoader.swift; sourceTree = "<group>"; };
 		7345ACFF2703D9470020AFB1 /* DaysUntilBirthdayForPod (iOS).app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = "DaysUntilBirthdayForPod (iOS).app"; sourceTree = BUILT_PRODUCTS_DIR; };
 		7345AD022703D9470020AFB1 /* DaysUntilBirthday.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DaysUntilBirthday.swift; sourceTree = "<group>"; };
 		7345AD042703D9470020AFB1 /* ContentView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContentView.swift; sourceTree = "<group>"; };
@@ -218,6 +222,7 @@
 			children = (
 				7345AD162703D9C30020AFB1 /* AuthenticationViewModel.swift */,
 				736F49BB270E102C00580053 /* BirthdayViewModel.swift */,
+				641495092C3E094B00C9A613 /* VerifiedAgeViewModel.swift */,
 			);
 			path = ViewModels;
 			sourceTree = "<group>";
@@ -226,6 +231,7 @@
 			isa = PBXGroup;
 			children = (
 				7345AD172703D9C30020AFB1 /* GoogleSignInAuthenticator.swift */,
+				6414950B2C3E0C7A00C9A613 /* VerificationLoader.swift */,
 				736F49B9270E05E200580053 /* BirthdayLoader.swift */,
 				7345AD192703D9C30020AFB1 /* UserProfileImageLoader.swift */,
 			);
@@ -421,9 +427,11 @@
 				739FCC46270E467600C92042 /* BirthdayView.swift in Sources */,
 				7345AD1B2703D9C30020AFB1 /* SignInView.swift in Sources */,
 				7345AD212703D9C30020AFB1 /* GoogleSignInAuthenticator.swift in Sources */,
+				6414950A2C3E094B00C9A613 /* VerifiedAgeViewModel.swift in Sources */,
 				7345AD232703D9C30020AFB1 /* UserProfileImageLoader.swift in Sources */,
 				7345AD1E2703D9C30020AFB1 /* UserProfileImageView.swift in Sources */,
 				736F49BC270E102C00580053 /* BirthdayViewModel.swift in Sources */,
+				6414950C2C3E0C7A00C9A613 /* VerificationLoader.swift in Sources */,
 				736F49BA270E05E200580053 /* BirthdayLoader.swift in Sources */,
 				7345AD242703D9C30020AFB1 /* UserProfileView.swift in Sources */,
 				7345AD202703D9C30020AFB1 /* AuthenticationViewModel.swift in Sources */,
--- a/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
+++ b/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
@@ -9,7 +9,6 @@
 /* Begin PBXBuildFile section */
 		4C6A2BA2321434ACBD2AF201 /* Pods_DaysUntilBirthdayForPod__macOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 926A15D393D684C68E09FB0F /* Pods_DaysUntilBirthdayForPod__macOS_.framework */; };
 		641495072C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
-		641495082C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
 		6414950A2C3E094B00C9A613 /* VerifiedAgeViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495092C3E094B00C9A613 /* VerifiedAgeViewModel.swift */; };
 		6414950C2C3E0C7A00C9A613 /* VerificationLoader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6414950B2C3E0C7A00C9A613 /* VerificationLoader.swift */; };
 		7345AD032703D9470020AFB1 /* DaysUntilBirthday.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7345AD022703D9470020AFB1 /* DaysUntilBirthday.swift */; };
--- a/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
+++ b/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
@@ -3,11 +3,13 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 51;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXBuildFile section */
 		4C6A2BA2321434ACBD2AF201 /* Pods_DaysUntilBirthdayForPod__macOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 926A15D393D684C68E09FB0F /* Pods_DaysUntilBirthdayForPod__macOS_.framework */; };
+		641495072C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
+		641495082C3C90E100C9A613 /* VerificationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 641495062C3C90E100C9A613 /* VerificationView.swift */; };
 		7345AD032703D9470020AFB1 /* DaysUntilBirthday.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7345AD022703D9470020AFB1 /* DaysUntilBirthday.swift */; };
 		7345AD052703D9470020AFB1 /* ContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7345AD042703D9470020AFB1 /* ContentView.swift */; };
 		7345AD072703D9480020AFB1 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 7345AD062703D9480020AFB1 /* Assets.xcassets */; };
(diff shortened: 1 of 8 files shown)


Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
```diff
@@ -577,14 +583,15 @@
...
+				DEVELOPMENT_TEAM = JA85M8JD4Z;
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = iOS/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.google.DaysUntilBirthday;
+				PRODUCT_BUNDLE_IDENTIFIER = "testing-com.google.DaysUntilBirthday";
```
mdmathias, line 594: Revert

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
```diff
@@ -577,14 +583,15 @@
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_ASSET_PATHS = "\"iOS/Preview Content\"";
+				DEVELOPMENT_TEAM = JA85M8JD4Z;
```
mdmathias, line 586: Remove

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
```diff
@@ -599,14 +606,15 @@
...
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = iOS/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.google.DaysUntilBirthday;
+				PRODUCT_BUNDLE_IDENTIFIER = "testing-com.google.DaysUntilBirthday";
```
mdmathias, line 617: Revert

Samples/Swift/DaysUntilBirthday/Shared/Services/GoogleSignInAuthenticator.swift
```diff
@@ -27,6 +27,32 @@ final class GoogleSignInAuthenticator: ObservableObject {
...
+
+    guard let rootViewController = UIApplication.shared.windows.first?.rootViewController else {
+      print("There is no root view controller!")
+      return
+    }
+
+    let verifyAccountDetail = GIDVerifyAccountDetail()
+    verifyAccountDetail.verifyAccountDetails(accountDetails, presenting: rootViewController) {
+      verifyResult, error in
+            guard let verifyResult = verifyResult else {
```
mdmathias, line 46: Fix indentation

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -22,6 +22,13 @@ final class AuthenticationViewModel: ObservableObject {
   /// The user's log in status.
   /// - note: This will publish updates when its value changes.
   @Published var state: State
+
+#if os(iOS)
+  /// The user's account verification status.
+  /// - note: This will publish updates when its value changes.
+  @Published var verificationState: VerificationState
```
mdmathias, line 29: I think it might be cleaner to create a `VerifiedAgeViewModel` that will hold this state. This will keep the `AuthenticationViewModel` tightly coupled to its view, which will do the same for `VerificationView`. Otherwise, `AuthenticationViewModel` will have some stuff that is only relevant to separate views, making it an "uber" view model.

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
```diff
@@ -0,0 +1,62 @@
...
+      .navigationTitle("Verified Account!")
+    case .unverified:
+      ProgressView()
+        .navigationTitle(NSLocalizedString("Unverified account",
+                                           comment: "Unverified account label"))
+    }
+  }
+
+  func formatDateWithDateFormatter(_ date: Date) -> String {
+      let dateFormatter = DateFormatter()
```
mdmathias, line 57: Fix indentation

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+import GoogleSignIn
+
+struct VerificationView: View {
+  @ObservedObject var authViewModel: AuthenticationViewModel
```
mdmathias, line 21: Will need to update when you have a `VerifiedAgeViewModel`.

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/Services/GoogleSignInAuthenticator.swift
brnnmrls: Done.

brnnmrls: Moved to VerificationLoader.swift

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
brnnmrls: Makes sense, went ahead and created the new view model.

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+import GoogleSignIn
+
+#if os(iOS)
+
+final class VerifiedAgeViewModel: ObservableObject {
```
mdmathias, line 22: Please add a doc comment for this class.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+  /// The user's account verification status.
+  /// - note: This will publish updates when its value changes.
+  @Published var verificationState: VerificationState
+
+  private var loader: VerificationLoader {
```
mdmathias, line 27: This will make a new `loader` every time it is called. Is that necessary? My sense is that you can instantiate once in the initializer. 

I know that `AuthenticationViewModel` does this, but that could be a mistake or for some unrelated reason; so, I figured I'd ask. :)

mdmathias: I guess part of the issue is that `GoogleSignInAuthenticator` and this loader pass themselves into the respective initializers. That means we can't create instances of these loaders in the view model initializers. I wonder if you could make these `lazy var`s? E.g.,

```
private lazy var loader: VerificationLoader = {
  return VerificationLoader(verifiedViewAgeMode: self)
}()
```

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+  /// Creates an instance of this view model.
+  init() {
+    self.verificationState = .unverified
+  }
+
+  /// Verifies the user.
+  func verifyAccountDetails() {
+    switch self.verificationState {
+    case .unverified:
+      loader.verifyAccountDetails()
```
mdmathias, line 40: What is the role of this `switch`? I recognize that the user cannot literally change, but knowledge of the user's verified age can. I.e., perhaps that user disconnected the app from this scope. In this case, if the app is still in memory, then it would still reflect that the user is verified, but that is based on older knowledge.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
```diff
@@ -583,15 +591,14 @@
...
-				DEVELOPMENT_TEAM = JA85M8JD4Z;
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = iOS/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "testing-com.google.DaysUntilBirthday";
+				PRODUCT_BUNDLE_IDENTIFIER = com.google.DaysUntilBirthday;
```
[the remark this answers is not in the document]
brnnmrls: It looks like it doesn't have quotes on the [Main Branch](https://github.com/google/GoogleSignIn-iOS/blob/main/Samples/Swift/DaysUntilBirthday/DaysUntilBirthday.xcodeproj/project.pbxproj), but it probably doesn't make a difference if I add since `com.google.DaysUntilBirthdayUITests-iOS-` is in quotes. I can add if you like, let me know what you think.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -0,0 +1,57 @@
...
+  /// - parameter verifiedAgeViewModel: The view model to use to set verification status on.
+  init(verifiedViewAgeModel: VerifiedAgeViewModel) {
+    self.verifiedAgeViewModel = verifiedViewAgeModel
+  }
+
+  /// Verifies the user's age based upon the selected account.
+  /// - note: Successful calls to this will set the `verifiedAgeViewModel`'s `verificationState` property.
+  func verifyAccountDetails() {
+    let accountDetails: [GIDVerifiableAccountDetail] = [
+      GIDVerifiableAccountDetail(accountDetailType:.ageOver18)
```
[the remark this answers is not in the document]
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    guard let rootViewController = UIApplication.shared.windows.first?.rootViewController else {
+      print("There is no root view controller!")
+      return
+    }
+
+    let verifyAccountDetail = GIDVerifyAccountDetail()
+    verifyAccountDetail.verifyAccountDetails(accountDetails, presenting: rootViewController) {
+      verifyResult, error in
+      guard let verifyResult = verifyResult else {
```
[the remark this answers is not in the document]
brnnmrls: You're right, changed.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -0,0 +1,57 @@
...
+  private var verifiedAgeViewModel: VerifiedAgeViewModel
+
+  /// Creates an instance of this loader.
+  /// - parameter verifiedAgeViewModel: The view model to use to set verification status on.
+  init(verifiedViewAgeModel: VerifiedAgeViewModel) {
+    self.verifiedAgeViewModel = verifiedViewAgeModel
+  }
+
+  /// Verifies the user's age based upon the selected account.
+  /// - note: Successful calls to this will set the `verifiedAgeViewModel`'s `verificationState` property.
```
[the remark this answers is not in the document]
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Looks like making it a `lazy var` works. Did we want this change on `GoogleSignInAuthenticator` as well?

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: The `switch` is meant to prevent redundant verification checks so whenever an instance of `VerifiedAgeViewModel` is created then the user's state is set to `unverified` from the beginning. Though, I see the concern where the app should reflect the most up-to-date verification status when the user disconnected the app from the scope. In this case, then it would be helpful to add an additional check if the `verificationState` is labeled `verified`. If not, then we will force re-verification. Would it makes sense to make another request for the scopes?

brnnmrls: Though this also seems to apply for the sign-in flow. If I disconnect the app from the sign-in scopes, then the information is still there due to memory. Is this expected? Wondering if the verify flow should be making an extra request to verify that the scopes have changed or not. Is there a way to add a listener for this?

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayForPod.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
mdmathias: Ah no worries then!

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
mdmathias: Let's leave the other one as is for now. We can make the change there if there is a need.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
mdmathias: There is no push driven listener we can implement for changes in sign in (or scopes) state changed in the user's account settings page. The issue I'm seeing is that the verified age signal is being treated as fixed once it's been fetched, but it isn't necessarily.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+    return VerificationLoader(verifiedViewAgeModel: self)
+  }
+
+  /// Creates an instance of this view model.
+  init() {
+    self.verificationState = .unverified
+  }
+
+  /// Verifies the user.
+  func verifyAccountDetails() {
```
mdmathias, line 38: I think we should make this clear that this is verifying age.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -0,0 +1,57 @@
...
+  func verifyAccountDetails() {
```
mdmathias, line 35: This should specify that age is being verified.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Sounds good, changed it back by removing the switch and making the request whenever the view method is called.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Discussed offline regarding refreshing the token whenever it's expired or near expiration. Since the refresh API is not included in this PR I'll add this in a separate PR after https://github.com/google/GoogleSignIn-iOS/pull/453.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -32,7 +32,7 @@ final class VerificationLoader: ObservableObject {
   /// Verifies the user's age based upon the selected account.
```
mdmathias, line 32: We don't verify a user's precise age. I think we should update the doc comment and method name to indicate that this method verifies that a user's age is over 18. 

`verifyUserAgeOver18() { ... }`

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
```diff
@@ -57,7 +57,10 @@ struct UserProfileView: View {
           NavigationLink(NSLocalizedString("Verify My Age", comment: "Verify Age"),
                          destination: VerificationView(verifiedAgeViewModel: verifiedAgeViewModel)
           .onAppear {
-            verifiedAgeViewModel.verifyAccountDetails()
+            verifiedAgeViewModel.verifyUserAge()
+          }
+          .onDisappear {
+            verifiedAgeViewModel.verificationState = .unverified
```
mdmathias, line 63: What role does `verificationState` play now (or will play in future prs)?

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
brnnmrls: `verificationState` determines what's displayed in [`VerificationView`](https://github.com/google/GoogleSignIn-iOS/blob/aef9ff79668024f8a107c8204f205bb9c4601b72/Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift#L24). For example, if the account is verified then the result object properties are listed, otherwise the view tells the user the account is unverified.

Additionally, we want to avoid falsely reflecting that the user is verified if the app is still in memory and instead ensure `verifyUserAgeOver18()` is called each time the user taps on "Verify My Age" to make the token request, so setting `verificationState` to `.unverified` will satisfy these intentions.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
mdmathias: Following up from an IRL conversation. @brnnmrls clarified how `verifiedAgeViewModel` propagates its state changes upon verification. Resolving.

mdmathias (approved):
LGTM. Please make sure the integration tests pass prior to merging. Thanks!

PR #454, opened 2024-07-12
Update AppCheckCore dependency to v11

### LINKED ISSUE #444: Increment AppCheckCore dependency to 11
# Context

[AppCheckCore](https://github.com/google/app-check) is incrementing it's minimum iOS version to 12, like we will per https://github.com/google/GoogleSignIn-iOS/issues/443.

# To Do

Increment our AppCheckCore dependency in [Package.swift](https://github.com/google/GoogleSignIn-iOS/blob/main/Package.swift#L51) and [GoogleSignIn.podspec](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn.podspec#L36) to 11.

mdmathias:
GoogleUtilities also has to be updated to v8 because that is AppCheckCore's version for this dependency.

Fixes #444

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Services/Foo.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Services/Foo.swift
@@ -1,8 +1,0 @@
-//
-//  Foo.swift
-//  DaysUntilBirthday
-//
-//  Created by Matt Mathias on 7/11/24.
-//
-
-import Foundation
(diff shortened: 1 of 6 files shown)


Samples/Swift/DaysUntilBirthday/Shared/Services/Foo.swift
yalkhder, line 1: Remove?

mdmathias: Grrr argh. Thought I removed that. Thanks!

yalkhder (approved):

dav3yblaz3 (commented):
get this going for test please thanks team

PR #459, opened 2024-07-18
Use class `GIDToken` to handle tokens in `GIDVerifiedAccountDetailResult`.

brnnmrls:
Update token properties in `GIDVerifiedAccountDetailResult` to use `GIDToken`'s `tokenString` and `expirationDate` properties to handle access and refresh token information.

### DIFF
--- a/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
+++ b/GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
@@ -44,6 +44,11 @@ NS_ASSUME_NONNULL_BEGIN
   if (self) {
     _verifiedAccountDetails = accountDetails;
     _verifiedAuthState = authState;
+    _accessToken = 
+        [[GIDToken alloc] initWithTokenString:authState.lastTokenResponse.accessToken
+                               expirationDate:authState.lastTokenResponse.accessTokenExpirationDate];
+    _refreshToken = [[GIDToken alloc] initWithTokenString:authState.refreshToken
+                                           expirationDate:nil];
   }
   return self;
 }
@@ -71,6 +76,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)updateVerifiedDetailsWithTokenResponse:(nullable OIDTokenResponse *)tokenResponse {
   if (tokenResponse) {
+    _accessToken = [[GIDToken alloc] initWithTokenString:tokenResponse.accessToken
+                                          expirationDate:tokenResponse.accessTokenExpirationDate];
+
     NSArray<NSString *> *accountDetailsString =
         [OIDScopeUtilities scopesArrayWithString:tokenResponse.scope];
     NSMutableArray<GIDVerifiableAccountDetail *> *verifiedAccountDetails = [NSMutableArray array];
@@ -85,17 +93,6 @@ NS_ASSUME_NONNULL_BEGIN
   } else {
     _verifiedAccountDetails = @[];
   }
-}
-
-- (nullable GIDToken *)accessToken {
-  return [[GIDToken alloc]
-          initWithTokenString:self.verifiedAuthState.lastTokenResponse.accessToken
-               expirationDate:self.verifiedAuthState.lastTokenResponse.accessTokenExpirationDate];
-}
-
-- (nullable GIDToken *)refreshToken {
-  return [[GIDToken alloc] initWithTokenString:self.verifiedAuthState.refreshToken
-                                expirationDate:nil];
 }
 
 - (BOOL)isEqual:(id)object {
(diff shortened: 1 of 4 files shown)


GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
```diff
@@ -85,16 +87,15 @@ - (void)updateVerifiedDetailsWithTokenResponse:(nullable OIDTokenResponse *)toke
   }
 }
 
-- (nullable NSString *)accessTokenString {
-  return self.verifiedAuthState.lastTokenResponse.accessToken;
-}
-
-- (nullable NSString *)refreshTokenString {
-  return self.verifiedAuthState.refreshToken;
+- (nullable GIDToken *)accessToken {
```
mdmathias, line 90: This will create a new `GIDToken` every time it's called. I don't think we should do that. Why not just make this a property and create it upon instantiation (and update it upon refresh)?

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
```diff
@@ -85,16 +87,15 @@ - (void)updateVerifiedDetailsWithTokenResponse:(nullable OIDTokenResponse *)toke
...
+  return [[GIDToken alloc]
+          initWithTokenString:self.verifiedAuthState.lastTokenResponse.accessToken
+               expirationDate:self.verifiedAuthState.lastTokenResponse.accessTokenExpirationDate];
 }
 
-- (nullable NSDate *)expirationDate {
-  return self.verifiedAuthState.lastTokenResponse.accessTokenExpirationDate;
+- (nullable GIDToken *)refreshToken {
+  return [[GIDToken alloc] initWithTokenString:self.verifiedAuthState.refreshToken
```
mdmathias, line 97: Same here. I don't think we want to create a new token each time this getter is called.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiedAccountDetailResult.h
```diff
@@ -25,19 +25,18 @@
...
 
-/// The date when the access token expires.
-@property(nonatomic, readonly, nullable) NSDate *expirationDate;
-/// The access token string.
-@property(nonatomic, copy, readonly, nullable) NSString *accessTokenString;
-/// The refresh token string.
-@property(nonatomic, copy, readonly, nullable) NSString *refreshTokenString;
+/// The access token object holding the string and expiration.
+@property(nonatomic, copy, readonly, nullable) GIDToken *accessToken;
+/// The refresh token object holding the string and no expiration.
```
mdmathias, line 38: Nit: remove "no" so that we don't have to remember to update these docs if expiration is ever needed. Developers can infer no expiration if they get a token with nil in that property.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
brnnmrls: This makes sense, definitely don't want to create a new one every time. 

Though, in regards to refreshing, it looks like `GIDToken.tokenString` and `GIDToken.expirationDate` are [readOnly](https://github.com/google/GoogleSignIn-iOS/blob/f54b8c7aaaf70a2dfc372ac65d05714fa093cf5b/GoogleSignIn/Sources/Public/GoogleSignIn/GIDToken.h#L25-L28), so the current token will not be updated. Seems like we could create a new token on refresh or change the `readonly` to `readwrite`, but I understand that the latter is unlikely. What do you think?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiedAccountDetailResult.h
brnnmrls: Sounds good, done.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
mdmathias: We don't want these tokens to be `readwrite`. They should be `readonly`.

brnnmrls: Sounds good, I moved them to the initializer and replace the access token when refreshing.

GoogleSignIn/Sources/GIDVerifyAccountDetail/Implementations/GIDVerifiedAccountDetailResult.m
brnnmrls: Same changes added above. Created upon instantiation.

mdmathias (approved):

PR #460, opened 2024-07-20
Create method to fetch the age verification signal from API.

brnnmrls:
The following creates a method to be called from the sample app to fetch the age verification signal. Upon calling the method this will create a URL session with the retrieved access token. Then, the response will be decoded and retrieve any age verification signal. This verification signal will be used in a following PR to inform the user if their age is verified over 18.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
```diff
@@ -0,0 +1,70 @@
...
+  init(from decoder: Decoder) throws {
+    let container = try decoder.singleValueContainer()
+    self.signal = try container.decode(String.self)
+  }
+
+  init(signal: String) {
+    self.signal = signal
+  }
+
+  static var noVerificationSignal: Verification? {
```
mdmathias, line 31: Does this need to be an optional if you're always returning the same value? Separately, can this be `static let noVerificationSignal: Verification = Verification(signal: "No signal found")`?

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
brnnmrls: Oops, does not need to be optional. Additionally, looks like the suggestion works too, changed!

mdmathias (approved):

PR #461, opened 2024-07-20
Create button on Verification View to fetch age verification signal.

brnnmrls:
The following creates a button that when tapped will start the flow to fetch the age verification signal from the URL. This will use [VerificationLoader fetchAgeVerificationSignal:] in #460 to start the request. The `ageVerificationView` will utilize this verification signal to determine what the app will show next:
1. If the verification signal is "AGE_OVER_18_STANDARD", then the app will present a sheet informing the user their age over 18 is verified.
2. If the verification signal is anything else or does not exist, then the app will present an alert informing the user that their age is not verified over 18 and more information of what's stored in the age verification signal (i.e. "AGE_PENDING").

Video demonstration of success and fail runs: [recordings](https://drive.google.com/drive/folders/1gTBsFImSq-y_pKlQg4I478ELSyTr8LP2?usp=sharing)

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+struct Verification: Decodable {
+  let signal: String
+
+  init(from decoder: Decoder) throws {
+    let container = try decoder.singleValueContainer()
+    self.signal = try container.decode(String.self)
+  }
+
+  init(signal: String) {
+    self.signal = signal
+  }
+
+  static var noVerificationSignal: Verification? {
+    return Verification(signal: "No signal found")
+  }
+}
+
+struct VerificationResponse: Decodable {
+  let verifications: [Verification]
+  let firstVerification: Verification
+
+  init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    self.verifications = try container.decode([Verification].self, forKey:.ageVerificationResults)
+    guard let first = verifications.first else {
+      throw Error.noVerificationInResult
+    }
+    self.firstVerification = first
+  }
+}
+
+extension VerificationResponse {
+  enum CodingKeys: String, CodingKey {
+    case ageVerificationResults
+  }
+}
+
+extension VerificationResponse {
+  enum Error: Swift.Error {
+    case noVerificationInResult
+  }
+}
+
+/*
+ {
+   "name": "ageVerification",
+   "verificationId": "A verification id string",
+   "ageVerificationResults": [
+     "AGE_PENDING"
+   ]
+ }
+ */
--- a/Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
@@ -123,16 +123,19 @@ final class VerificationLoader: ObservableObject {
   func fetchAgeVerificationSignal(verifyResult: GIDVerifiedAccountDetailResult,
                                   completion: @escaping () -> Void) {
     self.verificationPublisher(verifyResult: verifyResult) { publisher in
-      self.cancellable = publisher.sink { sinkCompletion in
-        if case .failure(let error) = sinkCompletion {
-          self.verification = Verification.noVerificationSignal
-          print("Error retrieving age verification: \(error)")
+      self.cancellable = publisher.sink(
+        receiveCompletion: { sinkCompletion in
+          if case .failure(let error) = sinkCompletion {
+            self.verification = Verification.noVerificationSignal
+            print("Error retrieving age verification: \(error)")
+            completion()
+          }
+        },
+        receiveValue: { verification in
+          self.verification = verification
+          completion()
         }
-        completion()
-      } receiveValue: { verification in
-        self.verification = verification
-        completion()
-      }
+      )
     }
   }
 }
--- a/Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import Combine
 import Foundation
 import GoogleSignIn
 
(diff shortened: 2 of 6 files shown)


Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -32,12 +41,31 @@ final class VerifiedAgeViewModel: ObservableObject {
   /// Creates an instance of this view model.
   init() {
     self.verificationState = .unverified
+    self.ageVerificationSignal = "Signal Unavailable"
```
mdmathias, line 44: Can this be an enum with a `String` raw value?

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -32,12 +41,31 @@ final class VerifiedAgeViewModel: ObservableObject {
...
+    self.isShowingAgeVerificationSignal = false
```
mdmathias, line 45: I don't think this needs to be set in the initializer if you give it a default value.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -32,12 +41,31 @@ final class VerifiedAgeViewModel: ObservableObject {
...
     loader.verifyUserAgeOver18()
   }
+
+  /// Fetches the age verification signal representing whether the user's age is over 18 or pending.
+  func fetchAgeVerificationSignal(verifyResult: GIDVerifiedAccountDetailResult) {
+    loader.fetchAgeVerificationSignal(verifyResult: verifyResult) {
+      self.ageVerificationSignal =
+        self.loader.verification?.signal ?? "Error: No verification details found"
+
+      if (self.ageVerificationSignal == "AGE_OVER_18_STANDARD"){
```
mdmathias, line 60: For example, the enum case raw values can be things like "AGE_OVER_18_STANDARD", which would be used to get an instance of a case matching that name. It makes dealing with strings safer (i.e., more type based).

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
```diff
@@ -52,6 +62,16 @@ struct VerificationView: View {
                  action:{refresh(results: result)})
         }
       }
+      .sheet(isPresented: $verifiedAgeViewModel.isShowingAgeVerificationSignal) {
+          AgeVerificationResultView(ageVerificationSignal: verifiedAgeViewModel.ageVerificationSignal)
+      }
+      .alert(isPresented: $verifiedAgeViewModel.isShowingAgeVerificationAlert, content: {
```
mdmathias, line 68: I think this is deprecated: https://developer.apple.com/documentation/swiftui/view/alert(ispresented:content:)

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Made `enum AgeVerificationSignal` to hold this value.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Done.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
brnnmrls: Made enum AgeVerificationSignal to hold these values.

Samples/Swift/DaysUntilBirthday/Shared/Views/VerificationView.swift
brnnmrls: Updated.

Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
```diff
@@ -0,0 +1,70 @@
...
+}
+
+struct VerificationResponse: Decodable {
+  let verifications: [Verification]
+  let firstVerification: Verification
+
+  init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    self.verifications = try container.decode([Verification].self, forKey:.ageVerificationResults)
+    guard let first = verifications.first else {
```
mdmathias: I think my preference is to not have this initializer `throw` and just make `firstVerification` an optional.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -14,34 +14,97 @@
...
+      switch result {
+      case .success(let urlSession):
+        guard let request = self?.request else {
+          return completion(Fail(error:.couldNotCreateURLRequest).eraseToAnyPublisher())
+        }
+        let verificationPublisher = urlSession.dataTaskPublisher(for: request)
+          .tryMap { data, error -> Verification in
+            let decoder = JSONDecoder()
+            let verificationResponse = try decoder.decode(VerificationResponse.self, from: data)
+            return verificationResponse.firstVerification
```
mdmathias: What happens if the init fails on 78? Maybe that is caught by the `mapError` below? If so, then I think a `throw`ing initializer is good and you can disregard my comment above.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -53,6 +116,37 @@ final class VerificationLoader: ObservableObject {
...
+  private var cancellable: AnyCancellable?
+
+  /// Fetches the age verification signal of the current user.
+  func fetchAgeVerificationSignal(verifyResult: GIDVerifiedAccountDetailResult,
+                                  completion: @escaping () -> Void) {
+    self.verificationPublisher(verifyResult: verifyResult) { publisher in
+      self.cancellable = publisher.sink { sinkCompletion in
+        switch sinkCompletion {
+        case .finished:
+          break
```
mdmathias: Is a `break` technically needed?

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
brnnmrls: Tried this out and can confirm if the init fails on 78 then this is caught by the `mapError`.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
brnnmrls: I believe so. Otherwise, I receive error `'case' label in a 'switch' must have at least one executable statement`. I believe a `break` or `()` is needed to break out of the switch and indicate you intend for the case to do nothing.

mdmathias (changes requested):
Can you attach a short video showing how this flows looks?

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/VerifiedAgeViewModel.swift
```diff
@@ -24,20 +24,57 @@ final class VerifiedAgeViewModel: ObservableObject {
   /// The user's account verification status.
   /// - note: This will publish updates when its value changes.
   @Published var verificationState: VerificationState
+  /// The age verification signal telling whether the user's age is over 18 or pending.
+  /// - note: This will publish updates when its value changes.
+  @Published var ageVerificationSignal: String
```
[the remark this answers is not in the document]
brnnmrls: You're right 🧐. Made this type `AgeVerificationSignal `.

brnnmrls:
> Can you attach a short video showing how this flows looks?

added to description! lmk if this should be more private

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
brnnmrls: Ohh I see what you mean. I've switched this over to an if statement so it only focuses on the error case and confirmed it works.

Samples/Swift/DaysUntilBirthday/Shared/Services/VerificationLoader.swift
```diff
@@ -124,14 +124,11 @@ final class VerificationLoader: ObservableObject {
...
-        switch sinkCompletion {
-        case .finished:
-          break
-        case .failure(let error):
+        if case .failure(let error) = sinkCompletion {
           self.verification = Verification.noVerificationSignal
           print("Error retrieving age verification: \(error)")
-          completion()
         }
+        completion()
```
mdmathias, line 131: I see two completions here. Does that mean we will call `completion()` _both_ when a value is received and when the `sink` completes? If so, should we call `completion()` twice?

brnnmrls: In the case of a failure or succession case, completion should only be called once. Looks like in this case it's being called twice. I moved the the completion to be within the `.failure` check but I believe this method also works if I only have the `completion()` at line 131 and remove the one at line 134 (now line 136).

mdmathias (approved):

Samples/Swift/DaysUntilBirthday/Shared/Models/Verification.swift
brnnmrls: Resolving in respect to [this comment](https://github.com/google/GoogleSignIn-iOS/pull/461#discussion_r1699176341).

Issue #408, opened 2024-05-06
Undefined symbols when used as dependency of multiple targets (Swift Packages in Xcode)


cjwirth:
**Describe the bug**
When building the app, I get undefined symbol errors, saying that it can't find symbols that exist in AppAuthCore:

```
ld: Undefined symbols:
  _OBJC_CLASS_$_OIDAuthState, referenced from:
       in GoogleSignIn.o
  _OBJC_CLASS_$_OIDAuthorizationRequest, referenced from:
       in GoogleSignIn.o
  _OBJC_CLASS_$_OIDAuthorizationService, referenced from:
       in GoogleSignIn.o
  _OBJC_CLASS_$_OIDIDToken, referenced from:
       in GoogleSignIn.o
  _OBJC_CLASS_$_OIDServiceConfiguration, referenced from:
       in GoogleSignIn.o
  _OBJC_CLASS_$_OIDURLQueryComponent, referenced from:
       in GoogleSignIn.o
  _OIDOAuthErrorResponseErrorKey, referenced from:
      +[GIDEMMSupport handleTokenFetchEMMError:completion:] in GoogleSignIn.o
  _OIDOAuthTokenErrorDomain, referenced from:
      ___53-[GIDGoogleUser refreshTokensIfNeededWithCompletion:]_block_invoke.22 in GoogleSignIn.o
  _OIDResponseTypeCode, referenced from:
      -[GIDSignIn authenticateInteractivelyWithOptions:] in GoogleSignIn.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This only happens if GoogleSignIn is added as a dependency of multiple targets that end up in the resulting app. If GoogleSignIn is only included into _one_ target, it seems to be OK. But when it's included in multiple targets, I get the error above.

I have created a [reproducible sample](https://github.com/cjwirth/GoogleSignIn-iOS/tree/broken-integration) using the ObjC sample app provided by this repo. 

**To Reproduce**
1. Open the sample app in Xcode
2. Create another Framework target that the main app depends on
3. Make the new Framework target depend on GoogleSignIn
4. Build the app -- see the errors

Or checkout the [reproducible sample](https://github.com/cjwirth/GoogleSignIn-iOS/tree/broken-integration) that I put together and try building it.

The app is basically structured like this in terms of project and dependencies:

```
.
└── AppTarget
    ├── GoogleSignIn
    ├── OnboardingFramework
    │   └── GoogleSignIn
[...]

PR #470, opened 2024-09-02
Fix Undefined symbols issue

alvarhansen:
Add AppAuthCore as explicit dependency to GoogleSignIn target as there are symbols that it uses from that library.

It is needed where GoogleSignIn is used as dynamic library/framework. This is the case when GoogleSignIn is added to multiple targets in dependency graph where 2 dynamic libraries/frameworks exist. 

Example of it can be seen in project that @cjwirth provided in #408.

Fixes: https://github.com/google/GoogleSignIn-iOS/issues/408

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -71,6 +71,7 @@ let package = Package(
       name: "GoogleSignIn",
       dependencies: [
         .product(name: "AppAuth", package: "AppAuth"),
+        .product(name: "AppAuthCore", package: "AppAuth"),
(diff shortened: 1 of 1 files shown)


danibachar:
Would love to see this merged, this low level SDK is used by so many other Google related SDKs, which makes it hard, or even impossible to fork and consume separately

angelolloqui:
Any update on this? we are getting issues importing GSI in a project that uses it from different subprojects. I believe this could fix it.

yalkhder (approved):

PR #473, opened 2024-09-04
Throw an exception if verified age scope is passed in sign-in request through the add scopes flow.

brnnmrls:
If a developer passes verification age scope to addScopes method, this results in an error (invalid_request). Instead of having the developer run through documentation and figure out why it errors, we'll throw an exception to let the developer know to use `GIDVerifyAccountDetail` instead.

### DIFF
--- a/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
+++ b/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+/// A registry to manage restricted scopes and their associated handling classes to track scopes
+/// that require separate flows within an application.
+@interface GIDRestrictedScopesRegistry : NSObject
+
+/// A set of strings representing the restricted scopes.
+@property (nonatomic, strong, readonly) NSSet<NSString *> *restrictedScopes;
+
+/// A dictionary mapping restricted scopes to their corresponding handling classes.
+@property (nonatomic, strong, readonly) NSDictionary<NSString *, Class> *scopeToClassMapping;
+
+/// This designated initializer sets up the initial restricted scopes and their corresponding handling classes.
+///
+/// @return An initialized `GIDRestrictedScopesRegistry` instance
+- (instancetype)init;
+
+/// Checks if a given scope is restricted.
+///
+/// @param scope The scope to check.
+/// @return YES if the scope is restricted; otherwise, NO.
+- (BOOL)isScopeRestricted:(NSString *)scope;
+
+/// Retrieves a dictionary mapping restricted scopes to their handling classes within a given set of scopes.
+///
+/// @param scopes A set of scopes.
+/// @return A dictionary where restricted scopes found in the input set are mapped to their corresponding handling classes.
+///     If no restricted scopes are found, an empty dictionary is returned.
+- (NSDictionary<NSString *, Class> *)restrictedScopeToClassMappingInSet:(NSSet<NSString *> *)scopes;
+
+@end
--- a/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
+++ b/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#import <TargetConditionals.h>
+
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+
 #import <Foundation/Foundation.h>
 
 /// A registry to manage restricted scopes and their associated handling classes to track scopes
@@ -45,3 +49,5 @@
 - (NSDictionary<NSString *, Class> *)restrictedScopeToClassMappingInSet:(NSSet<NSString *> *)scopes;
 
 @end
+
+#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
--- a/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
+++ b/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
@@ -43,10 +43,10 @@
 
 /// Retrieves a dictionary mapping restricted scopes to their handling classes within a given set of scopes.
 ///
-/// @param scopes A set of scopes.
+/// @param scopes A set of scopes to lookup their handling class.
 /// @return A dictionary where restricted scopes found in the input set are mapped to their corresponding handling classes.
 ///     If no restricted scopes are found, an empty dictionary is returned.
-- (NSDictionary<NSString *, Class> *)restrictedScopeToClassMappingInSet:(NSSet<NSString *> *)scopes;
+- (NSDictionary<NSString *, Class> *)restrictedScopesToClassMappingInSet:(NSSet<NSString *> *)scopes;
 
 @end
 
--- a/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
+++ b/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
@@ -17,6 +17,8 @@
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifiableAccountDetail.h"
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDVerifyAccountDetail.h"
 
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+
 @implementation GIDRestrictedScopesRegistry
 
 - (instancetype)init {
@@ -46,3 +48,5 @@
 }
 
 @end
+
+#endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
--- a/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
+++ b/GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
@@ -1,0 +1,48 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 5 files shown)


brnnmrls:
Taking care of swift-button-functional-test.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -989,6 +997,16 @@ - (void)assertValidPresentingViewController {
   }
 }
 
+// Asserts the requested scopes are valid.
+- (void)assertValidScopes:(NSArray<NSString *> *)scopes {
+  if ([scopes containsObject:@"https://www.googleapis.com/auth/verified.age.over18.standard"]) {
```
mdmathias, line 1002: Could we instead make a set of "restricted scopes" and use that in a comparison to see if the provided scopes overlaps with the set of restricted scopes? If there is an overlap, the exception should say what the restricted scopes are and what should be done.

Or maybe you could create a dictionary that maps restricted scope to a class that should be used? For now, it'll just be the one scope, but it could grow overtime (e.g., if we add more verifiable account details that require their own flows). You could use that mapping to tell callers that they have supplied a scope that should use a different class/flow.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -257,6 +257,10 @@ - (void)addScopes:(NSArray<NSString *> *)scopes
                                                   addScopesFlow:YES
                                                      completion:completion];
 
+  // Explicitly throw an exception for invalid or restricted scopes in the request.
+  // This should be done in class GIDVerifyAccountDetail.
```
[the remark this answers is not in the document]
brnnmrls: Removed.

GoogleSignIn/Sources/GIDSignIn.m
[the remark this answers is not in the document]
brnnmrls: There is in GIDVerifiableAccountDetail.h! Imported the file and used the constant instead.

brnnmrls: Decided to follow through with the dictionary suggestion. New method `-[GIDSignIn assertValidScopes:]` will run through the requested scopes and log all scopes (and their classes) that may not be supported by the add scopes flow in GIDSignIn.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -997,15 +995,31 @@ - (void)assertValidPresentingViewController {
   }
 }
 
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 // Asserts the requested scopes are valid.
 - (void)assertValidScopes:(NSArray<NSString *> *)scopes {
-  if ([scopes containsObject:@"https://www.googleapis.com/auth/verified.age.over18.standard"]) {
+  NSDictionary *scopeToClassMapping = @{
```
mdmathias, line 1001: Defining this mapping here will mean it is recreated with each method call. That's not necessary. We can make this cleaner. I see a couple of ways forward.

1) Define this dictionary as a private property or as a constant in this file.

2) Create some kind of "restricted scopes registry" (e.g., `GIDRestrictedScopesRegistry`) that keeps track of restricted scopes (i.e., scopes that should not be passed here and require their own flows). `GIDRestrictedScopesRegistry` would hold a set of scopes whose use are restricted from being passed into `-[GIDSignIn addScopes...]` and therefore require special handling. `GIDRestrictedScopesRegistry` would have properties for the set of restricted scopes and a mapping from restricted scope to class. It would also have methods to i) determine if a supplied scope is restricted and ii) return the restricted scopes from a passed set of scopes. These two methods could also return something (a custom object?) that provides the class that should be used for the scope's flow.

Option 1 is faster. Option 2 is more scalable and testable. My preference is for option 2 since it isn't that much more work.

mdmathias: I could go for either though; so, it's up to you!

mdmathias (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -997,15 +995,31 @@ - (void)assertValidPresentingViewController {
   }
 }
 
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 // Asserts the requested scopes are valid.
```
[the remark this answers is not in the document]
brnnmrls: Done.

GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
```diff
@@ -0,0 +1,53 @@
...
+
+/// Checks if a given scope is restricted.
+///
+/// @param scope The scope to check.
+/// @return YES if the scope is restricted; otherwise, NO.
+- (BOOL)isScopeRestricted:(NSString *)scope;
+
+/// Retrieves a dictionary mapping restricted scopes to their handling classes within a given set of scopes.
+///
+/// @param scopes A set of scopes.
```
mdmathias, line 46: "A set of scopes to lookup their handling class."

GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
```diff
@@ -0,0 +1,52 @@
...
+    };
+  }
+  return self;
+}
+
+- (BOOL)isScopeRestricted:(NSString *)scope {
+  return [self.restrictedScopes containsObject:scope];
+}
+
+- (NSDictionary<NSString *, Class> *)restrictedScopeToClassMappingInSet:(NSSet<NSString *> *)scopes {
```
mdmathias, line 39: "scope" in the method name should be plural: `-restrictedScopes...`

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -996,24 +997,19 @@ - (void)assertValidPresentingViewController {
 }
 
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
-// Asserts the requested scopes are valid.
 - (void)assertValidScopes:(NSArray<NSString *> *)scopes {
-  NSDictionary *scopeToClassMapping = @{
-    kAccountDetailTypeAgeOver18Scope : [GIDVerifyAccountDetail class],
-  };
-  NSMutableString *errorMessage =
+  GIDRestrictedScopesRegistry *registry = [[GIDRestrictedScopesRegistry alloc] init];
```
mdmathias, line 1001: Probably better to make this a private property on `GIDSignIn`.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -996,24 +997,19 @@ - (void)assertValidPresentingViewController {
...
+  NSDictionary<NSString *, Class> *restrictedScopesMapping = [registry restrictedScopeToClassMappingInSet:[NSSet setWithArray:scopes]];
```
mdmathias, line 1002: Is this line too long? If so, please wrap following our style.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -996,24 +997,19 @@ - (void)assertValidPresentingViewController {
...
     [NSMutableString stringWithString:@"The following scopes are not supported in the 'addScopes' flow. "
                                        "Please use the appropriate classes to handle these:\n"];
-  Boolean unsupportedScopeFound = NO;
-
-  for (NSString *scope in scopes) {
-    Class scopeClass = scopeToClassMapping[scope];
-    if (scopeClass) {
-      unsupportedScopeFound = YES;
-      [errorMessage appendFormat:@"%@ -> %@\n", scope, NSStringFromClass(scopeClass)];
+    for (NSString *restrictedScope in restrictedScopesMapping) {
```
mdmathias, line 1008: You can enumerate keys and values at the same time: https://developer.apple.com/documentation/foundation/nsdictionary/1414570-enumeratekeysandobjectsusingbloc

Would remove the need for the line below.

GoogleSignIn/Tests/Unit/GIDRestrictedScopesRegistryTest.m
```diff
@@ -0,0 +1,44 @@
...
+@implementation GIDRestrictedScopesRegistryTest
+- (void)testIsScopeRestricted {
+    GIDRestrictedScopesRegistry *registry = [[GIDRestrictedScopesRegistry alloc] init];
+    BOOL isRestricted = [registry isScopeRestricted:kAccountDetailTypeAgeOver18Scope];
+    XCTAssertTrue(isRestricted);
+}
+
+- (void)testRestrictedScopeToClassMappingInSet {
+    GIDRestrictedScopesRegistry *registry = [[GIDRestrictedScopesRegistry alloc] init];
+    NSSet<NSString *> *scopes = [NSSet setWithObjects:kAccountDetailTypeAgeOver18Scope, @"some_other_scope", nil];
```
mdmathias, line 35: I think you've logically captured this in the tests below, but would you also please add a test ensuring the `"some_other_scope"` is not in the mapping below?

mdmathias (changes requested):

GoogleSignIn/Sources/GIDRestrictedScopesRegistry.h
brnnmrls: Done.

GoogleSignIn/Sources/GIDRestrictedScopesRegistry.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDSignIn.m
brnnmrls: Done.

GoogleSignIn/Sources/GIDSignIn.m
brnnmrls: Good catch! Formatted.

GoogleSignIn/Sources/GIDSignIn.m
brnnmrls: Cool! Incorporated this in, thanks.

GoogleSignIn/Tests/Unit/GIDRestrictedScopesRegistryTest.m
brnnmrls: Added! Asserted that `"some_other_scope"` isn't in the mapping.

mdmathias (changes requested):

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -998,18 +1000,19 @@ - (void)assertValidPresentingViewController {
 
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 - (void)assertValidScopes:(NSArray<NSString *> *)scopes {
-  GIDRestrictedScopesRegistry *registry = [[GIDRestrictedScopesRegistry alloc] init];
-  NSDictionary<NSString *, Class> *restrictedScopesMapping = [registry restrictedScopeToClassMappingInSet:[NSSet setWithArray:scopes]];
+  _registry = [[GIDRestrictedScopesRegistry alloc] init];
```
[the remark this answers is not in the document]
brnnmrls: Added.

mdmathias (approved):

PR #478, opened 2024-09-27
Remove unification of requested and granted scopes in add scopes flow.

brnnmrls:
With VwG age scope being incorporated, our add scopes flow is running into an "Authorization error: invalid_request". This is happening since `grantedScopes` includes the VwG age scope and is being passed into the request. A solution we'll incorporate is removing the unification to only pass in specified `requestedScopes` into the request.

### DIFF
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -283,15 +283,9 @@ static NSString *const kClientAssertionTypeParameterValue =
     }
     return;
   }
-
-  // Use the union of granted and requested scopes.
-  if ([grantedScopes containsObject:kAccountDetailTypeAgeOver18Scope] && grantedScopes.count > 1) {
-    options.scopes = [requestedScopes allObjects];
-  } else {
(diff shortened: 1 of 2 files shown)


GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -283,11 +283,9 @@ - (void)addScopes:(NSArray<NSString *> *)scopes
     }
     return;
   }
-
-  // Use the union of granted and requested scopes.
-  [grantedScopes unionSet:requestedScopes];
-  options.scopes = [grantedScopes allObjects];
-
+  
+  options.scopes = [requestedScopes allObjects];
```
mdmathias, line 287: Maybe add comment here that says that we pass `include_granted_scopes=true` here so the previously granted scopes will still be there.

mdmathias (approved):

GoogleSignIn/Sources/GIDSignIn.m
brnnmrls: Done.

PR #512, opened 2025-04-10
Remove requirement of a current user in the VwG flow.

brnnmrls:
Remove the requirement of a current user to allow developers to use VwG without a signed in account.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS.swift
+++ b/Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS.swift
@@ -26,6 +26,7 @@ class DaysUntilBirthdayUITests_iOS: XCTestCase {
   private let additionalAccessHeaderText = "Days Until Birthday wants additional access to your Google Account"
   private let appTrustWarningText = "Make sure you trust Days Until Birthday"
   private let chooseAnAccountHeaderText = "Choose an account"
+  private let consentRequestHeaderText = "Share some info with Days Until Birthday"
   private let notNowText = "Not Now"
   private let timeout: TimeInterval = 30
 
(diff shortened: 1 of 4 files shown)


Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS.swift
```diff
@@ -68,6 +69,23 @@ class DaysUntilBirthdayUITests_iOS: XCTestCase {
       return XCTFail("Signing out should return user to sign in view")
     }
   }
+
+  func testFetchVerificationSignalAndDisconnect() {
+    sampleApp.launch()
+
+    XCTAssertTrue(signIn())
```
mdmathias, line 76: Is this needed? Maybe the sample app has the sign in screen first? I think that's the case...anyway, is it possible to test this flow without having a sign in first? That would get things closer to the change in this pr.

mdmathias: One way to accomplish this is to have a verify button under the sign in button so that the verify flow can be tested without a sign in first.

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS.swift
```diff
@@ -275,6 +371,17 @@ extension DaysUntilBirthdayUITests_iOS {
     }
   }
 
+  /// Proceeds through the view with header "Share some info with Days Until Birthday" if needed.
```
mdmathias, line 374: Nit: spacing

mdmathias (changes requested):

Samples/Swift/DaysUntilBirthday/DaysUntilBirthdayUITests(iOS)/DaysUntilBirthdayUITests_iOS.swift
brnnmrls: Done.

PR #516, opened 2025-04-25
Update scorecards.yml to use upload-artifact@v4.6.2

camden-king:
V3 is deprecated as of November 30, 2024 https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact.

SKIP_INTEGRATION_TESTS=YES


brnnmrls:
Found out that swift-button-functional-test is failing, but is not related to this change. Skipping the test for now, but this will need a fix in the future.

brnnmrls (approved):

PR #544, opened 2025-08-12
Support Swift6 version to Birthday sample app.

khetiSubrat:
Support Swift6 version to Birthday sample app.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Services/UserProfileImageLoader.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Services/UserProfileImageLoader.swift
@@ -22,38 +22,50 @@ typealias GIDImage = NSImage
 
 import Combine
 import SwiftUI
-import GoogleSignIn
+@preconcurrency import GoogleSignIn
 
 /// An observable class for loading the current user's profile image.
-final class UserProfileImageLoader: ObservableObject {
+@MainActor final class UserProfileImageLoader: ObservableObject {
   private let userProfile: GIDProfileData
-  private let imageLoaderQueue = DispatchQueue(label: "com.google.days-until-birthday")
+  private var imageLoadingTask: Task<Void, Never>?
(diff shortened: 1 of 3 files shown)


Samples/Swift/DaysUntilBirthday/Shared/Services/UserProfileImageLoader.swift
```diff
@@ -22,10 +22,10 @@ typealias GIDImage = NSImage
...
 import Combine
 import SwiftUI
-import GoogleSignIn
+@preconcurrency import GoogleSignIn
 
 /// An observable class for loading the current user's profile image.
-final class UserProfileImageLoader: ObservableObject {
+@MainActor final class UserProfileImageLoader: ObservableObject {
   private let userProfile: GIDProfileData
   private let imageLoaderQueue = DispatchQueue(label: "com.google.days-until-birthday")
```
w-goog: Please change this to store the Task you created on line 42, and cancel the task in the deinit.

```suggestion
  private var imageLoadingTask: Task<Void, Never>?
```

And then:
```
    // In the init, assign the task to the property
    imageLoadingTask = Task {
        await loadProfileImage()
    }  
    // Cancel in the deinit
    deinit {
        imageLoadingTask?.cancel() // Cancel the task on deinit
    }

w-goog (changes requested):
Thank you for updating this! One small change requested, then we should be good to go :)

(Note: We'll explicitly support Swift 6 in the repo soon, at which point some of these changes may be reverted)

Samples/Swift/DaysUntilBirthday/Shared/Services/UserProfileImageLoader.swift
khetiSubrat: Thanks for quick response. Done the changes as suggested.

w-goog (approved):
Tested locally by pulling in these changes, and they look good. Thank you for making them! :)

PR #546, opened 2025-08-20
Update `swift-tools` version to use Swift 6 when available.

w-goog:
Updating the `swift-tools` version will let the compiler choose Swift 6 by default. Providing Package@swift-5.5.swift supports clients still on Xcode 15.x. See [Apple docs](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/swiftversionspecificpackaging) for manifest versioning details.


w-goog:
Rerunning pod lint, as the failure appears to be a transient issue with file caching.

PR #550, opened 2025-09-03
GIDTokenClaim Implementation + Unit Tests

AkshatG6:
This pull request introduces the `GIDTokenClaim` class, which allows for requesting specific OIDC claims for an ID token.

Key changes:
*   Adds the `GIDTokenClaim` class to represent a single OIDC claim.
*   Includes factory methods for creating essential and non-essential `auth_time` claims.
*   Provides unit tests to validate the implementation.

### DIFF
--- a/GoogleSignIn/Sources/GIDTokenClaim.m
+++ b/GoogleSignIn/Sources/GIDTokenClaim.m
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h"
+
+// Private interface to declare the internal initializer
+@interface GIDTokenClaim ()
+
+- (instancetype)initWithName:(NSString *)name
+                   essential:(BOOL)essential NS_DESIGNATED_INITIALIZER;
+
+@end
+
+@implementation GIDTokenClaim
+
+// Private designated initializer
+- (instancetype)initWithName:(NSString *)name essential:(BOOL)essential {
+  self = [super init];
+  if (self) {
+    _name = [name copy];
+    _essential = essential;
+  }
+  return self;
+}
+
+#pragma mark - Factory Methods
+
++ (instancetype)authTimeClaim {
+  return [[self alloc] initWithName:@"auth_time" essential:NO];
+}
+
++ (instancetype)essentialAuthTimeClaim {
+  return [[self alloc] initWithName:@"auth_time" essential:YES];
+}
+
+#pragma mark - NSObject
+
+- (BOOL)isEqual:(id)object {
+  // 1. Check if the other object is the same instance in memory.
+  if (self == object) {
+    return YES;
+  }
+
+  // 2. Check if the other object is not a GIDTokenClaim instance.
+  if (![object isKindOfClass:[GIDTokenClaim class]]) {
+    return NO;
+  }
+
+  // 3. Compare the properties that define equality.
+  GIDTokenClaim *other = (GIDTokenClaim *)object;
+  return [self.name isEqualToString:other.name] &&
+         self.isEssential == other.isEssential;
+}
+
+- (NSUInteger)hash {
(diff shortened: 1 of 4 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h
```diff
@@ -0,0 +1,46 @@
...
+/// Whether the claim is requested as essential.
+@property (nonatomic, readonly, getter=isEssential) BOOL essential;
+
+// Making initializers unavailable to force use of factory methods.
+- (instancetype)init NS_UNAVAILABLE;
+
+#pragma mark - Factory Methods
+
+/// Creates a *non-essential* (voluntary) "auth_time" claim object.
++ (instancetype)authTimeClaim;
```
mdmathias, line 40: Do you think it would be helpful for developers to add a `- Note:` that links to the OIDC docs on essential claims. I can see how it'd be nice to share here because not all developers may be aware; however, the docs are a little confusing, and I don't know if linking would just lead to more confusion. What do you think?

GoogleSignIn/Tests/Unit/GIDTokenClaimTest.m
```diff
@@ -0,0 +1,48 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+#import <XCTest/XCTest.h>
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h"
+@interface GIDTokenClaimTest : XCTestCase
```
mdmathias, line 19: Nit: newline above

GoogleSignIn/Tests/Unit/GIDTokenClaimTest.m
```diff
@@ -0,0 +1,48 @@
...
+@end
+
+@implementation GIDTokenClaimTest
+
+- (void)testAuthTimeClaim_PropertiesAreCorrect {
+  GIDTokenClaim *claim = [GIDTokenClaim authTimeClaim];
+  XCTAssertEqualObjects(claim.name, @"auth_time");
```
mdmathias: Perhaps useful to add string constants to `GIDTokenClaim.h`: `kAuthTimeClaim`. You could then compare to that here.

mdmathias (changes requested):

GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h
AkshatG6: I agree on both the points. Maybe, we can reference the [SiWG documentation](https://developers.google.com/identity/sign-in/ios/start-integrating), which will have a simpler explanation (and there we also add the link to OIDC docs?). 

IMO, either way we should add the note referencing one of the docs. Let me know which one do you prefer.

GoogleSignIn/Tests/Unit/GIDTokenClaimTest.m
AkshatG6: Fixed.

GoogleSignIn/Tests/Unit/GIDTokenClaimTest.m
AkshatG6: I've added the string constant.

GoogleSignIn/Tests/Unit/GIDTokenClaimTest.m
```diff
@@ -15,20 +15,23 @@
 
 #import <XCTest/XCTest.h>
 #import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h"
+
+static NSString *const kAuthTimeClaimName = @"auth_time";
```
mdmathias: This should be declared in `GIDTokenClaim.h`, implemented and used in `GIDTokenClaim.m`

GoogleSignIn/Sources/GIDTokenClaim.m
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "GoogleSignIn/Sources/Public/GoogleSignIn/GIDTokenClaim.h"
+
+NSString * const kAuthTimeClaimName = @"auth_time";
```
mdmathias, line 19: You should follow this pattern:

https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h#L33

https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDSignIn.m#L98

mdmathias: Please disregard!

PR #553, opened 2025-09-10
Updated GIDSignIn + GIDSignInInternalOptions Implementations + Unit Tests

AkshatG6:
This pull request updates the existing implementation of `GIDSignIn` class and `GIDSignInInternalOptions` class.

Key changes:

* Updates the `signWith...` initializers to provide public api's to request for tokenClaims.
* Updates `GIDSignInInternalOptions` to store tokenClaims.
* Updates the sign-in flow to include tokenClaims within the `OIDAuthorizationRequest`.
* Provides unit tests to validate the implementation.

### DIFF
--- a/GoogleSignIn/Sources/GIDSignIn.m
+++ b/GoogleSignIn/Sources/GIDSignIn.m
@@ -28,6 +28,7 @@
 #import "GoogleSignIn/Sources/GIDCallbackQueue.h"
 #import "GoogleSignIn/Sources/GIDScopes.h"
 #import "GoogleSignIn/Sources/GIDSignInCallbackSchemes.h"
+#import "GoogleSignIn/Sources/GIDTokenClaimsInternalOptions.h"
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 #import <AppCheckCore/GACAppCheckToken.h>
 #import "GoogleSignIn/Sources/GIDAppCheck/Implementations/GIDAppCheck.h"
@@ -136,6 +137,9 @@ static NSString *const kIncludeGrantedScopesParameter = @"include_granted_scopes
 static NSString *const kLoginHintParameter = @"login_hint";
 static NSString *const kHostedDomainParameter = @"hd";
 
+// Parameter for requesting the token claims.
+static NSString *const kTokenClaimsParameter = @"claims";
+
 // Parameters for auth and token exchange endpoints using App Attest.
 static NSString *const kClientAssertionParameter = @"client_assertion";
 static NSString *const kClientAssertionTypeParameter = @"client_assertion_type";
@@ -169,6 +173,7 @@ static NSString *const kConfigOpenIDRealmKey = @"GIDOpenIDRealm";
   // set when a sign-in flow is begun via |signInWithOptions:| when the options passed don't
   // represent a sign in continuation.
   GIDSignInInternalOptions *_currentOptions;
+  GIDTokenClaimsInternalOptions *_tokenClaimsInternalOptions;
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
   GIDAppCheck *_appCheck API_AVAILABLE(ios(14));
 #endif // TARGET_OS_IOS && !TARGET_OS_MACCATALYST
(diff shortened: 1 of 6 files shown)


GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -225,6 +226,99 @@ NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.")
...
     NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.");
 
+/// Starts an interactive sign-in flow on iOS using the provided tokenClaims.
+///
+/// The completion will be called at the end of this process.  Any saved sign-in state will be
+/// replaced by the result of this flow.  Note that this method should not be called when the app is
+/// starting up, (e.g in `application:didFinishLaunchingWithOptions:`); instead use the
+/// `restorePreviousSignInWithCompletion:` method to restore a previous sign-in.
+///
+/// @param presentingViewController The view controller used to present `SFSafariViewController` on
```
mdmathias: For this param doc comment and all below: `SFSafariViewController` is not relevant. iOS 9 and 10 really aren't either. Regardless, the logic of which VC to use is not controlled here (it's in AppAuth). Please change here and everywhere else in this file to "The view controller used to present the authorization flow."

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -225,6 +226,99 @@ NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.")
...
+/// @param completion The optional block that is called on completion.  This block will
+///     be called asynchronously on the main queue.
+- (void)signInWithPresentingViewController:(UIViewController *)presentingViewController
+                               tokenClaims:(nullable NSSet<GIDTokenClaim *> *)tokenClaims
+                                completion:
+    (nullable void (^)(GIDSignInResult *_Nullable signInResult,
+                       NSError *_Nullable error))completion
+    NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.");
+
+/// Starts an interactive sign-in flow on iOS using the provided hint, and tokenClaims.
```
mdmathias: Nit: remove comma.

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
```diff
@@ -225,6 +226,99 @@ NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.")
...
+- (void)signInWithPresentingViewController:(UIViewController *)presentingViewController
+                                      hint:(nullable NSString *)hint
+                          additionalScopes:(nullable NSArray<NSString *> *)additionalScopes
+                               tokenClaims:(nullable NSSet<GIDTokenClaim *> *)tokenClaims
+                                completion:
+    (nullable void (^)(GIDSignInResult *_Nullable signInResult,
+                       NSError *_Nullable error))completion
+    NS_EXTENSION_UNAVAILABLE("The sign-in flow is not supported in App Extensions.");
+
+/// Starts an interactive sign-in flow on iOS using the provided hint, additional scopes, nonce
```
mdmathias: Nit: comma after `nonce`.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -1149,6 +1267,24 @@ - (void)assertValidPresentingViewController {
   }
 }
 
+- (BOOL)processTokenClaimsForOptions:(GIDSignInInternalOptions *)options
+                               error:(NSError **)error {
+  if (!options.tokenClaims) {
+    return YES; // Success
```
mdmathias: Remove comments like "// Success" as the BOOL return conveys the same idea.

GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -1149,6 +1267,24 @@ - (void)assertValidPresentingViewController {
   }
 }
 
+- (BOOL)processTokenClaimsForOptions:(GIDSignInInternalOptions *)options
```
mdmathias: You may be following an existing pattern in this file, but this method "works" by relying on side effect behavior, which is not great. The side effect, setting the token claims JSON on the passed options, is separated from the return, a BOOL. IMHO side effects like this make code harder to read (e.g., in the first snippet pasted below).

So, you're code above would go from:

```
// If tokenClaims are invalid or JSON serialization fails, return with an error.
if (![self processTokenClaimsForOptions:options error:&claimsError]) {
  if (options.completion) {
    self->_currentOptions = nil;
    dispatch_async(dispatch_get_main_queue(), ^{
      options.completion(nil, claimsError);
    });
  }
  return;
}
```

To this:

```
// If tokenClaims are invalid or JSON serialization fails, return with an error.
options.tokenClaimsAsJSON = [self processTokenClaimsForOptions:options error:&claimsError])
if (options.completion && /* Check that token claims are good on options; otherwise, set completion with error */) {
  self->_currentOptions = nil;
  dispatch_async(dispatch_get_main_queue(), ^{
    options.completion(nil, claimsError);
  });
  return;
}
```

Wouldn't it be simpler to just return the token claims json?

GoogleSignIn/Tests/Unit/GIDSignInInternalOptionsTest.m
```diff
@@ -63,6 +64,56 @@ - (void)testDefaultOptions {
...
+  id presentingViewController = OCMStrictClassMock([UIViewController class]);
+#elif TARGET_OS_OSX
+  id presentingWindow = OCMStrictClassMock([NSWindow class]);
+#endif // TARGET_OS_IOS || TARGET_OS_MACCATALYST
+  NSString *loginHint = @"login_hint";
+  NSArray<NSString *> *scopes = @[@"scope1", @"scope2"];
+  NSString *nonce = @"test_nonce";
+  NSSet<GIDTokenClaim *> *tokenClaims = [NSSet setWithObject:[GIDTokenClaim authTimeClaim]];
+
+  // The expected scopes array will contain the provided scopes plus the default profile scopes.
```
mdmathias: Nit: remove comment

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -8,6 +8,10 @@
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR COhttp://www.apache.org/licenses/LICENSE-2.0
```
mdmathias: Nit: fix/revert

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -1533,10 +1632,18 @@ - (void)OAuthLoginWithAddScopesFlow:(BOOL)addScopesFlow
...
                            additionalScopes:nil
                                       nonce:nonce
+                                tokenClaims:tokenClaims
                                  completion:completion];
       }
     }
 
+    if (tokenClaimsError) {
+      XCTestExpectation *tokenClaimsErrorExpectation =
+          [self expectationWithDescription:@"Callback called"];
```
mdmathias: Let's not reuse "Callback called" since it's extremely vague. Please use something more descriptive like "Token claims error expectation."

GoogleSignIn/Tests/Unit/GIDSignInTest.m
```diff
@@ -1533,10 +1632,18 @@ - (void)OAuthLoginWithAddScopesFlow:(BOOL)addScopesFlow
                                        hint:_hint
                            additionalScopes:nil
                                       nonce:nonce
+                                tokenClaims:tokenClaims
                                  completion:completion];
       }
     }
 
+    if (tokenClaimsError) {
+      XCTestExpectation *tokenClaimsErrorExpectation =
```
mdmathias: I'm a little confused by this...you make an expectation and then fulfill it right away? Can you help me to understand what is going on here?

GoogleSignIn/Sources/Public/GoogleSignIn/GIDSignIn.h
AkshatG6: Done.

GoogleSignIn/Sources/GIDSignIn.m
AkshatG6: That's right. I have updated the function accordingly.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
AkshatG6: Removed this expectation altogether.

GoogleSignIn/Tests/Unit/GIDSignInTest.m
AkshatG6: I have updated this condition.

GoogleSignIn/Sources/GIDSignIn.m
AkshatG6: I don't think we even need `processTokenClaimsForOptions`.

PR #555, opened 2025-09-15
Updated sample app to support auth_time

AkshatG6:
This pull request improves the sign-in flow to request, decode, and display the `auth_time` claim for both iOS and macOS sample apps. 

### Key changes:
- Request auth_time claim: Updated the `GIDSignIn.sharedInstance.signIn(...)` call to include `GIDTokenClaim.authTime()`.
- JWT Decoding: Extended `AuthenticationViewModel` to parse the `idToken` from the `GIDGoogleUser` and extract the `auth_time` value.
- UI Update: The user's profile view has been updated to display the formatted "Last sign-in date" if it's available.

### UI Comparison

---

#### iOS

| Before | After |
| :---: | :---: |
| image | image  |
---
#### macOS

| Before | After |
| :---: | :---: |
| image | image |

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
@@ -27,7 +27,7 @@ final class AuthenticationViewModel: ObservableObject {
   }
 
   /// The user's `auth_time` as found in `idToken`.
-  /// - note: If the user is logged out, then this will default to nil.
+  /// - note: If the user is logged out, then this will default to `nil`.
   var authTime: Date? {
     switch state {
     case .signedIn(let user):


Samples/Swift/DaysUntilBirthday/Shared/DaysUntilBirthday.swift
```diff
@@ -29,6 +29,7 @@ struct DaysUntilBirthday: App {
           GIDSignIn.sharedInstance.restorePreviousSignIn { user, error in
             if let user = user {
               self.authViewModel.state = .signedIn(user)
+              self.authViewModel.authTime = UserDefaults.standard.object(forKey: "authTime") as? Date
```
khetiSubrat: I noticed that authTime is currently being stored in UserDefaults. Since this value is related to authentication, have you considered using the Keychain for storage instead?

Samples/Swift/DaysUntilBirthday/Shared/Services/GoogleSignInAuthenticator.swift
```diff
@@ -36,12 +36,14 @@ final class GoogleSignInAuthenticator: ObservableObject {
       return
     }
     let manualNonce = UUID().uuidString
+    let tokenClaims: Set<GIDTokenClaim> = Set([GIDTokenClaim.authTime()])
```
khetiSubrat: Consider declaring tokenClaims once at the top of the signIn() function instead of separately for iOS and macOS. This will reduce duplication.

Samples/Swift/DaysUntilBirthday/Shared/DaysUntilBirthday.swift
AkshatG6: Updated.

Samples/Swift/DaysUntilBirthday/Shared/Services/GoogleSignInAuthenticator.swift
AkshatG6: Done.

Samples/Swift/DaysUntilBirthday/Shared/DaysUntilBirthday.swift
mdmathias: I _think_ the `user` here should have an `idToken` on [this property](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/Public/GoogleSignIn/GIDGoogleUser.h#L60). I.e., we don't need to set `authTime` at all. It could just be a computed property on the view model.

Samples/Swift/DaysUntilBirthday/Shared/Services/GoogleSignInAuthenticator.swift
```diff
@@ -66,7 +68,10 @@ final class GoogleSignInAuthenticator: ObservableObject {
       return
     }
 
-    GIDSignIn.sharedInstance.signIn(withPresenting: presentingWindow) { signInResult, error in
+    GIDSignIn.sharedInstance.signIn(
+      withPresenting: presentingWindow,
+      tokenClaims: tokenClaims
+      ) { signInResult, error in
```
mdmathias: I think you need to delete two spaces here to align the parenthesis.

Samples/Swift/DaysUntilBirthday/Shared/DaysUntilBirthday.swift
AkshatG6: Thanks for the input @mdmathias and @khetiSubrat. 

I've updated the code to rely on computing `authTime` directly from `idToken` as suggested.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -69,7 +81,49 @@ final class AuthenticationViewModel: ObservableObject {
   @MainActor func addBirthdayReadScope(completion: @escaping () -> Void) {
       authenticator.addBirthdayReadScope(completion: completion)
   }
+  
+  var formattedAuthTimeString: String? {
```
mdmathias, line 86: Nit: please fix formatting.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -25,6 +25,18 @@ final class AuthenticationViewModel: ObservableObject {
   private var authenticator: GoogleSignInAuthenticator {
     return GoogleSignInAuthenticator(authViewModel: self)
   }
+
+  // 3. Change authTime to a computed property
```
mdmathias: Nit: No need for this comment

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -69,7 +81,49 @@ final class AuthenticationViewModel: ObservableObject {
...
+  var formattedAuthTimeString: String? {
+          guard let date = authTime else { return nil }
+          let formatter = DateFormatter()
+          formatter.dateFormat = "MM/dd/yyyy"
+          return formatter.string(from: date)
+  }
+}
+
+private extension AuthenticationViewModel {
+    func decodeAuthTime(fromJWT jwt: String) -> Date? {
```
mdmathias: Please follow the spacing we use throughout GSI.

mdmathias: I think we use 2 instead of 4 spaces.

Samples/Swift/DaysUntilBirthday/Shared/ViewModels/AuthenticationViewModel.swift
```diff
@@ -69,7 +81,49 @@ final class AuthenticationViewModel: ObservableObject {
   @MainActor func addBirthdayReadScope(completion: @escaping () -> Void) {
       authenticator.addBirthdayReadScope(completion: completion)
   }
+  
+  var formattedAuthTimeString: String? {
+          guard let date = authTime else { return nil }
+          let formatter = DateFormatter()
+          formatter.dateFormat = "MM/dd/yyyy"
```
mdmathias: Can you add timestamp? `auth_time` is more specific than Month Day Year.

mdmathias (changes requested):

mdmathias (approved):

PR #559, opened 2025-10-01
Ensure `additionalTokenRefreshParametersForAuthSession` returns a string dictionary

brnnmrls:
This PR addresses a runtime crash that occurs when the dictionary returned by `additionalTokenRefreshParametersForAuthSession:` contains non-string values, leading to a type mismatch in GTMAppAuth.

This fix is implemented in `GIDEMMSupport` and makes the Google Sign-In library more robust by:
- Iterating through the additionalParameters dictionary using the `enumerateKeysAndObjectsUsingBlock:` instance method
- Checking the type of each value and explicitly converting `NSNumber` values to `NSString` 
- Safely ignoring any other non-string types to prevent future crashes.

This ensures that the dictionary returned to GTMAppAuth is always of the correct type.

### DIFF
--- a/GoogleSignIn/Sources/GIDEMMSupport.m
+++ b/GoogleSignIn/Sources/GIDEMMSupport.m
@@ -120,17 +120,54 @@ additionalTokenRefreshParametersForAuthSession:(GTMAuthSession *)authSession {
       [GIDEMMSupport updatedEMMParametersWithParameters:authSession.authState.lastTokenResponse
                                                             .additionalParameters];
 
-  // Ensure returned dictionary has values of only type String.
-  NSMutableDictionary<NSString *, NSString *> *convertedParameters =
-      [NSMutableDictionary dictionary];
-  [originalParameters enumerateKeysAndObjectsUsingBlock:^(NSString *key, id value, BOOL *stop) {
-    if ([value isKindOfClass:[NSNumber class]]) {
-      convertedParameters[key] = [value stringValue];
-    } else if ([value isKindOfClass:[NSString class]]) {
-      convertedParameters[key] = value;
+  NSMutableDictionary *updatedParameters = [NSMutableDictionary dictionary];
+  // Will first make any json objects within our original parameters into a string.
+  [updatedParameters enumerateKeysAndObjectsUsingBlock:^(NSString *key, id value, BOOL *stop) {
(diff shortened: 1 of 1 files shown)


GoogleSignIn/Sources/GIDEMMSupport.m
```diff
@@ -116,8 +116,21 @@ + (NSDictionary *)parametersWithParameters:(NSDictionary *)parameters
...
-          authSession.authState.lastTokenResponse.additionalParameters];
+  NSDictionary *originalParameters =
+      [GIDEMMSupport updatedEMMParametersWithParameters:authSession.authState.lastTokenResponse
+                                                            .additionalParameters];
+
+  // Ensure returned dictionary has values of only type String.
+  NSMutableDictionary<NSString *, NSString *> *convertedParameters =
+      [NSMutableDictionary dictionary];
+  [originalParameters enumerateKeysAndObjectsUsingBlock:^(NSString *key, id value, BOOL *stop) {
+    if ([value isKindOfClass:[NSNumber class]]) {
```
mdmathias: I think we want to be a bit more complete than checking for an `NSNumber`. We can leverage `NSJSONSerialization` to help us out.

1. Turn the `originalParameters` dictionary into data via [+[NSJSONSerialization dataWithJSONObject:options:error:]](https://developer.apple.com/documentation/foundation/jsonserialization/data(withjsonobject:options:)?language=objc)
2. Take that `NSData` and make an `NSString` from it: [-[NSString initWithData:encoding:]](https://developer.apple.com/documentation/foundation/nsstring/init(data:encoding:)?language=objc)
3. Now that you have a string dictionary, make that `NSData` via [[-[NSString dataUsingEncoding:]](https://developer.apple.com/documentation/foundation/nsstring/data(using:)?language=objc)
4. Pass that `NSData` to [+[NSJSONSerialization JSONObjectWithData:options:error:]](https://developer.apple.com/documentation/foundation/jsonserialization/jsonobject(with:options:)-8demi?language=objc)

The resulting JSON object _should_ be a `NSDictionary<NSString *, NSString *> *`. The idea behind step 2 is that it should make everything in the `NSData` an `NSString`.

I recommend writing some unit tests with various flavors of types in the values of `originalParameters`.

PR #561, opened 2025-10-07
Ensure `additionalTokenRefreshParametersForAuthSession` returns a string dictionary

AkshatG6:
This PR addresses a runtime crash that occurs when the dictionary returned by `additionalTokenRefreshParametersForAuthSession:` contains non-string values, leading to a type mismatch in GTMAppAuth.

This fix is implemented in `GIDEMMSupport` by iterating through the `additionalParameters` dictionary and converts values to strings with the following logic :-

- `NSString` values: Preserves them as-is.
- `NSNumber` values: Converts booleans to their string representation ("true" or "false") and all other numbers to their standard string value.
- `NSArray` and `NSDictionary` values: Serializes them into a single, compact `JSON` string.

This ensures that the dictionary returned to GTMAppAuth is always of the correct type.

### DIFF
--- a/GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
@@ -59,6 +59,11 @@ static NSString *const kEMMKey = @"emm_support";
 static NSString *const kDeviceOSKey = @"device_os";
 static NSString *const kEMMPasscodeInfoKey = @"emm_passcode_info";
 
+@interface GIDEMMSupport (Private)
++ (NSDictionary<NSString *, NSString *> *)
+    dictionaryWithStringValuesFromDictionary:(NSDictionary *)originalDictionary;
+@end
+
 @interface GIDEMMSupportTest : XCTestCase
   // The view controller that has been presented, if any.
 @property(nonatomic, strong, nullable) UIViewController *presentedViewController;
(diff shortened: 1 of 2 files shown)


GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -274,6 +279,80 @@ - (void)testHandleTokenFetchEMMError_errorIsNotEMM {
   [self waitForExpectations:@[ called ] timeout:1];
 }
 
+# pragma mark - String Conversion Tests
+
+- (void)testStringConversion_withAnyNumber_isConvertedToString {
+  NSDictionary *inputDictionary = @{ @"number_key": @12345 };
+
+  NSDictionary *resultDictionary = [GIDEMMSupport
```
mdmathias: Maybe just put `[GIDEMMSupport` on the next line too.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -59,6 +59,11 @@
 static NSString *const kDeviceOSKey = @"device_os";
 static NSString *const kEMMPasscodeInfoKey = @"emm_passcode_info";
 
+@interface GIDEMMSupport (Private)
```
mdmathias: I don't think you need to directly test this new method. I think the idea is that we want to make sure that the existing calls to `GIDEMMSupport` in `GIDSignIn` produce the additional parameter dictionaries that we expect (i.e., instances of `NSDictionary` that do not crash the GTMAppAuth delegate method call).

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -274,6 +279,80 @@ - (void)testHandleTokenFetchEMMError_errorIsNotEMM {
...
+  XCTAssertTrue([resultDictionary[@"number_key"] isKindOfClass:[NSString class]],
+                @"The value should be an NSString.");
+  XCTAssertEqualObjects(resultDictionary[@"number_key"], @"12345",
+                        @"The NSNumber should be converted to a string.");
+}
+
+- (void)testStringConversion_withNumberOne_isConvertedToString {
+  NSDictionary *inputDictionary = @{ @"number_key": @1 };
+
+  NSDictionary *resultDictionary = [GIDEMMSupport
```
mdmathias: Same comment about styling.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
```diff
@@ -274,6 +279,80 @@ - (void)testHandleTokenFetchEMMError_errorIsNotEMM {
...
+  XCTAssertTrue([resultDictionary[@"number_key"] isKindOfClass:[NSString class]],
+                @"The value should be an NSString.");
+  XCTAssertEqualObjects(resultDictionary[@"number_key"], @"0",
+                        @"The NSNumber should be converted to a string.");
+}
+
+- (void)testStringConversion_withBooleanYes_isConvertedToTrueString {
+  NSDictionary *inputDictionary = @{ @"bool_key": @YES };
+
+  NSDictionary *resultDictionary = [GIDEMMSupport
```
mdmathias: Same comment re styling.

GoogleSignIn/Tests/Unit/GIDEMMSupportTest.m
AkshatG6: I have updated the tests to instead test the `parametersWithParameters...` method.

mdmathias (approved):

PR #567, opened 2025-12-11
Remove keychain entries if fresh install

mdmathias:


README.md
camden-king: Can we include similar iOS information about it needing to be first here as well.

camden-king: Can you replace to with "as the first item in"

PR #569, opened 2026-01-06
Add support for `claims` like `auth_time`

AkshatG6:
This PR acts as a main feature branch that adds support for requesting `claims` via the `SignInWith...` public method. Currently, only `auth_time` claim is supported.


brnnmrls (approved):
Please make sure the unit tests are passing before merging.

Issue #565, opened 2025-12-02
Refresh_token_expires_in not getting from google OAuth token api by passing ServerAuthCode.


anuragverse14:
**Describe the bug**
I am getting a serverAuthCode which doesn't gives Refresh_token_expires_in from google Oauth token api call when user not checkmarked the ongoing access for google account.

**To Reproduce**
1. Call GIDSignIn.sharedInstance.signIn(withPresenting: rootViewController, hint: email, additionalScopes: scopes)
2. User get prompted for google sign inand then scopes list will be shown to the user.
3. User will **not select Ongoing access** and select all required scopes, then click on continue.
4. Code side we will get serverAuthCode and we will use google Oauth token api to get refresh token.
5. We will get refresh token but not refresh_token_expires_in to know whether refresh token is long lived or not.

**Expected behavior**
As user has not selected ongoing access, we should be getting refresh_token_expires_in and it should be 15551999.

**Environment**
- Device: iPhone 11
- OS: 18.2.1
- Browser: Internal browser(may be safari) launch by GIDSignIn

**Additional context**
Either you can expose refresh_token_expires_in from GIDSignIn response or can provide a serverAuthCode which helps in getting the same. I have tried with serverAuthCode got from android app and call google OAuth token api and that gives the refresh_token_expires_in.

anuragverse14:
I am looking for the same, but it is not accessible outside the library. 
One more thing i can't pass access_type=offline in GIDSignIn.sharedInstance.signIn(.....) function, so i can't get offline refresh token. So I need to call googleapi.com/oauth api along with serverAuthCode for getting a refresh token with offline access. Actually i need refersh_token_expires_in value in that api call response, but i never get there. 

Best way is to pass access_type=offline in GIDSignIn.sharedInstance.signIn(.....)  call and expose refresh_token_expires_in outside the library(Public access). This way i can get refresh token and refresh_token_expires_in.

Could you please update library functions with above requirement in next version?

w-goog:
Re: the expiration date - mea culpa! We did not add that when a handful of Google APIs added support for refresh token rotation. That's being added in the PR linked above, and we'll close this issue once it's complete.

[...]

PR #577, opened 2026-02-04
Expose refresh token expiration date.

brnnmrls:
In our token response from the server, we receive a "refresh_token_expires_in" value. This PR adds handling for this value and checks for refresh token expiration in refreshTokensIfNeeded.

Fixes [#565](https://github.com/google/GoogleSignIn-iOS/issues/565)

### DIFF
--- a/GoogleSignIn/Tests/Unit/GIDSignInTest.m
+++ b/GoogleSignIn/Tests/Unit/GIDSignInTest.m
@@ -310,6 +310,7 @@ static NSString *const kNonEssentialAuthTimeClaimsJsonString =
   OCMStub([_authState alloc]).andReturn(_authState);
   OCMStub([_authState initWithAuthorizationResponse:OCMOCK_ANY]).andReturn(_authState);
   _tokenResponse = OCMStrictClassMock([OIDTokenResponse class]);
+  OCMStub([_tokenResponse additionalParameters]).andReturn(@{});
   _tokenRequest = OCMStrictClassMock([OIDTokenRequest class]);
   _authorization = OCMStrictClassMock([GTMAuthSession class]);
   _keychainStore = OCMStrictClassMock([GTMKeychainStore class]);


GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -459,6 +459,25 @@ - (void)testRefreshTokensIfNeededWithCompletion_handleConcurrentRefresh {
   [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testRefreshTokensIfNeededWithCompletion_noRefresh_givenRefreshTokenExpired {
+  NSTimeInterval expiresIn = -10;
```
w-goog, line 463: Can you make the other tokens not expired, so that we're testing only the specific refresh token behavior?

w-goog (approved):
Approved pending the one requested change :)

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
brnnmrls: If I do this then this would not reach the refresh token logic. This is due to the fact that we check if either IdToken and/or AccessToken are expired first ([ref](https://github.com/google/GoogleSignIn-iOS/blob/913b4005ea26aebe1c97d54e35ad82a515924c71/GoogleSignIn/Sources/GIDGoogleUser.m#L121-L127)). If both of these tokens aren't expired, then we immediately return them to the client thus not reaching the refreshToken check that comes after.

The test where both of the tokens are not expired is already made [here](https://github.com/google/GoogleSignIn-iOS/blob/913b4005ea26aebe1c97d54e35ad82a515924c71/GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m#L365-L389). Although this doesn't have the refreshToken expiration, it should produce the same output. I could add the expired refreshToken to this test if we want to be safe.

Separately, I could add another test where if only one of the tokens are expired, then we should still continue to the refreshToken logic.

Let me know what you think.

w-goog: Ah! All good then, thank you for explaining :)

PR #591, opened 2026-03-05
Move GIDSignInButton import outside of conditional

freak4pc:
GIDSignInButton already gates internally for the proper OS, and gating in the umbrella header is causing build warnings in Xcode

Supersedes #342


freak4pc:
Hey @w-goog - I opened this PR myself since the other one is from 2023. Would appreciate a merge, I signed the CLA a while back.

Thanks!

w-goog (approved):
Thank you! Love getting a warning cleaned up :) I've checked and it doesn't look like SPM or pods includes this explicitly in header search paths, so this should be good to go.

freak4pc:
Any chance to get a release cut with this fix?

PR #597, opened 2026-04-01
Gandhiakshat/sbc amr sample app

AkshatG6:
This PR updates the sample app to display all `claims` in a collective, navigable list rather than showing them individually in the user profile.

### DIFF
--- a/Samples/Swift/DaysUntilBirthday/Shared/Models/Claim.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Models/Claim.swift
@@ -1,0 +1,6 @@
+
+struct Claim: Identifiable {
+  let key: String
+  let value: String
+  var id: String {key}
+}
--- a/Samples/Swift/DaysUntilBirthday/Shared/Models/Claim.swift
+++ b/Samples/Swift/DaysUntilBirthday/Shared/Models/Claim.swift
@@ -20,5 +20,5 @@ import Foundation
 struct Claim: Identifiable {
   let key: String
   let value: String
-  var id: String {key}
+  var id: String { key }
 }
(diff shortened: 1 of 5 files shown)


Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
```diff
@@ -17,6 +17,27 @@
 import SwiftUI
 import GoogleSignIn
 
+/// A view that displays a list of ID token claims.
+struct ClaimsListView: View {
+  let claims: [(key: String, value: String)]
```
mdmathias: Why not just make a `Claim` type with `key` and `value` properties?

mdmathias: Or, can you use `GIDClaim`?

mdmathias (changes requested):
Can you add a video or screenshot?

Samples/Swift/DaysUntilBirthday/iOS/UserProfileView.swift
AkshatG6: Creating a new `Claim` type because `GIDClaim` already exists and is used for requests instead of responses.

AkshatG6:
> Can you add a video or screenshot?

image

image

Samples/Swift/DaysUntilBirthday/Shared/Models/Claim.swift
```diff
@@ -0,0 +1,6 @@
+
+struct Claim: Identifiable {
+  let key: String
+  let value: String
+  var id: String {key}
```
mdmathias: Nit: Add space before and after `key`.

mdmathias (approved):

PR #604, opened 2026-06-12
g-orchestrated: add Claim.swift to DaysUntilBirthdayForPod project

w-goog:
When DaysUntilBirthday is built w/Cocoapods, it isn't able to find Claim.swift - this PR adds it to the appropriate project file.


AkshatG6 (approved):

w-goog:
Folded into #603 — the Claim.swift fix rides along with the AppAuth 2.1 bump so both land together.

PR #608, opened 2026-07-23
Fix crash in `additionalTokenRefreshParameters` when `emmSupport` is `nil`

AkshatG6:
Overview
This PR fixes an edge case missed by the previous fix for the Swift casting crash in `additionalTokenRefreshParameters` (#564).

The Problem
The previous fix correctly converted non-string values (like `NSNumber`) into their `NSString` representations to avoid a Swift type mismatch exception. However, it missed the scenario where `emmSupport` is `nil` or `NO`. In that case, `[GIDEMMSupport parametersWithParameters:emmSupport:isPasscodeInfoRequired:]` returned the original parameters dictionary early, bypassing the newly added string conversion helper.

The Fix
This change updates the early return block to apply `[GIDEMMSupport dictionaryWithStringValuesFromDictionary:parameters]`, ensuring that the dictionary is safely stringified regardless of the `emmSupport` configuration.

### DIFF
--- a/GoogleSignIn/Sources/GIDEMMSupport.m
+++ b/GoogleSignIn/Sources/GIDEMMSupport.m
@@ -95,7 +95,7 @@ typedef NS_ENUM(NSInteger, ErrorCode) {
                                                        emmSupport:(nullable NSString *)emmSupport
                                            isPasscodeInfoRequired:(BOOL)isPasscodeInfoRequired {
   if (!emmSupport) {
-    return parameters;
+    return [GIDEMMSupport dictionaryWithStringValuesFromDictionary:parameters];
   }
   NSMutableDictionary *allParameters = [(parameters ?: @{}) mutableCopy];
   allParameters[kEMMSupportParameterName] = emmSupport;
--- a/GoogleSignIn/Sources/GIDEMMSupport.h
+++ b/GoogleSignIn/Sources/GIDEMMSupport.h
@@ -38,6 +38,8 @@ NS_ASSUME_NONNULL_BEGIN
     (NSDictionary *)parameters;
 
 /// Gets a new set of URL parameters that also contains EMM-related URL parameters if needed.
+///
+/// @note This function only returns parameter keys in the `NSDictionary` for string/number/bools.
 + (NSDictionary<NSString *,NSString *> *)parametersWithParameters:(NSDictionary *)parameters
                                                        emmSupport:(nullable NSString *)emmSupport
                                            isPasscodeInfoRequired:(BOOL)isPasscodeInfoRequired;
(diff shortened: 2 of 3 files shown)


GoogleSignIn/Sources/GIDEMMSupport.m
w-goog, line 94: Please add a comment in the header file stating that this function only returns parameter keys in the NSDictionary for string/number/bools.

AkshatG6: Done. I've added the comment in the header file.

w-goog, line 94: Let's also fix the tests - I believe the tests always try emmSupport = true, so we're not testing all combinations of parameters :) Ensure that a few variations of NSDictionary are tested; both emmSupport true/false is tested (also, why is this parameter a string????), and isPasscodeInfoRequired true/false is tested.

w-goog (changes requested):
Requested test changes :)

GoogleSignIn/Sources/GIDEMMSupport.m
AkshatG6: I've added a couple of tests to ensure we are testing all the combinations. The existing tests should cover the rest.

Regarding your question about why `emmSupport` is a string. It represents the [EMM protocol version](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDSignIn.m#L789), which is currently defined as a string constant [(kEMMVersion = @"1")](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDSignIn.m#L96) in GIDSignIn. Because of this, emmSupport can only ever be "1" (when the device is on iOS 9+) or nil (for older Mac Catalyst). 

A false value isn't a valid state for this parameter, so testing for nil and "1" should cover all the possible scenarios.

w-goog: The signature accepts any NSString - so while you are logically correct, the function has (theoretically) defined behaviour if `emmSupport` is `false` or `goose` - it should return the parameters unchanged. So, we should either a) test that behaviour or b) only allow valid input :)

w-goog (changes requested):

GoogleSignIn/Sources/GIDEMMSupport.m
AkshatG6: Going with option a). 

I've added a new test for `ArbitraryEMMSupport` case.

w-goog (approved):
This is good to merge :) You can override the swift integration test to do so. Before merging, please file a follow-up issue to make the following changes:

1) That the `parametersWithParameters` function drops key parameters that contain values like a nested array.
2) We should fix the test swizzling UIDevice to Not Do That (or ensure all the tests are doing it the same way; using the real UIDevice will be non-hermetic).
3) Consider upleveling these tests to the public API.

AkshatG6:
Merging this PR.

1. Created https://github.com/google/GoogleSignIn-iOS/issues/611 to track the requested `GIDEMMSupport` refactoring, UIDevice swizzling fixes, and upleveling of the unit tests.

2. As discussed, I am overriding the failing Swift integration test to merge this PR. The failure is due to the iOS 18 simulators breaking the hardcoded XCUITest scripts, which will be fixed in a follow-up PR.

PR #623, opened 2026-08-14
Add wrapperIdentifier for callers to self-id in logs

w-goog:
A small number of clients bundle GoogleSignIn-iOS with their own SDKs. To ensure that the way we understand adoption of the SDK is accurate, we'd like to allow those bundlers to say "Hi, this clientID may be App123, but I, Bundler0, am handling it on their behalf".

This is entirely optional and represents no functional change to the way authorization requests are handled on the backend.


w-goog:
Oops, this should land later, and not be merged into the refactor.

PR #625, opened 2026-08-21
Feature/add sdkwrapper identifier

w-goog:
A small number of clients bundle GoogleSignIn-iOS with their own SDKs. To ensure that the way we understand adoption of the SDK is accurate, we'd like to allow those bundlers to say "Hi, this clientID may be App123, but I, Bundler0, am handling it on their behalf".

This is entirely optional and represents no functional change to the way authorization requests are handled on the backend.


GoogleSignIn/Sources/GIDSignIn.m
```diff
@@ -649,6 +649,14 @@ + (GIDSignIn *)sharedInstance {
   return sharedInstance;
 }
 
++ (nullable NSString *)wrapperIdentifier {
+  return [GIDSignInPreferences wrapperIdentifier];
+}
+
++ (void)setWrapperIdentifier:(nullable NSString *)wrapperIdentifier {
+  [GIDSignInPreferences setWrapperIdentifier:wrapperIdentifier];
```
mdmathias, line 657: Do we want to enforce the "set once" language in property's doc comment?

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
```diff
@@ -478,6 +484,95 @@ - (void)testRefreshTokensIfNeededWithCompletion_noRefresh_givenRefreshTokenExpir
   [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testWrapperIdentifier_PresentOnRefreshRequestWhenSet {
+  GIDSignIn.wrapperIdentifier = @"firebase";
+
+  // Both tokens expired 10 seconds ago.
+  GIDGoogleUser *user = [self googleUserWithAccessTokenExpiresIn:-10 idTokenExpiresIn:-10];
+
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Callback is called"];
```
mdmathias: Can we make the description more descriptive? I know the pattern in the library is to use "Callback is called," but I've had trouble in past with debugging failed tests when I see a wall of that output. It's hard to find _which_ callback wasn't called.

GoogleSignIn/Sources/GIDSignInPreferences.m
```diff
@@ -79,11 +117,55 @@ + (NSString *)environment {
...
+  if (sanitized == nil) {
+    NSLog(@"[Google Sign-In iOS]: the SDK wrapper identifier '%@' was rejected, because it must be "
+          "non-empty and contain only printable ASCII characters (U+0020 to U+007E).",
+          wrapperIdentifier);
+    return;
+  }
+
+  os_unfair_lock_lock(&gWrapperIdentifierLock);
+  NSString *current = gWrapperIdentifier;
+  if (current != nil && ![current isEqualToString:sanitized]) {
```
mdmathias, line 143: Ah, the "set once" enforcement is here. That's a little surprising, but I think I understand the reasoning. I don't know if this can be addressed, but I do still feel like it's confusing to see the doc comment on the property of the other class, and then for the actual enforcement to be here. ¯_(ツ)_/¯

Probably no big deal. Feel free to close.

mdmathias (approved):

GoogleSignIn/Sources/GIDSignInPreferences.m
w-goog: I added a comment here as well :) The enforcement being here is largely just due to the property / "implementation" split.

GoogleSignIn/Sources/GIDSignIn.m
w-goog: Added a comment in the other location to indicate the referenced doc comment.

GoogleSignIn/Tests/Unit/GIDGoogleUserTest.m
w-goog: Done :) I've updated all three to describe the expected result.

PR #626, opened 2026-08-21
[App Check] Preliminary test of App Check 12 (Swift implementation) integration

paulb777:
Integration testing for https://github.com/google/app-check/pull/111. 

There's no intention to release CocoaPods again, but the port is included here for increased test coverage.

We'd ideally like to move to a minimum of iOS 15, but iOS 13 would be ok.  iOS 12 won't work because of the reliance of async/await in the new implementation.

TODO:
- [ ] Remove CocoaPods changes
- [ ] Restore GHA workflows
- [ ] Wait for Firebase 13 to release
- [ ] Update Package.swift to specify AppCheck 11 and 12


w-goog:
Our next release (hopefully out by the end of this week) will move to 15! :)

paulb777:
Oh great! I'll update this PR after that.   We're still several weeks from the AppCheck update.

cc: @ncooke3

GoogleSignIn/Sources/GIDAppCheck/Implementations/Fake/GIDAppCheckProviderFake.h
```diff
@@ -16,7 +16,7 @@
 #import <Foundation/Foundation.h>
 
 #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
-#import <AppCheckCore/GACAppCheckProvider.h>
+@import AppCheckCore;
```
ncooke3, line 19: To investigate more:
```suggestion
#if __has_include(<AppCheckCore/GACAppCheckProvider.h>)
#import <AppCheckCore/GACAppCheckProvider.h> // Supports AppCheck v11
#else
@import AppCheckCore; // Supports AppCheck v12
#endif  // __has_include(<AppCheckCore/GACAppCheckProvider.h>)
```

paulb777: As long as we stick to SPM, `@import AppCheckCore;` works fine with both AppCheck 11 and 12.

GoogleSignIn.podspec
```diff
@@ -33,7 +33,7 @@ The Google Sign-In SDK allows users to sign in with their Google account from th
   ]
   s.ios.framework = 'UIKit'
   s.osx.framework = 'AppKit'
-  s.dependency 'AppCheckCore', '~> 11.0'
+  s.dependency 'AppCheckCore', '~> 12.0'
```
ncooke3: Before merging. Revert so CocoaPods is only compatible with AppCheckCore 11.x
```suggestion
  s.dependency 'AppCheckCore', '~> 11.0'
```

Samples/ObjC/SignInSample/Podfile
ncooke3: Revert when ready to merge

PR #628, opened 2026-08-27
Update GTMAppAuth to 6.0.0

w-goog:
GTMAppAuth needs to be updated for GSI to compile on Xcode 27. This comes with a few additional updates:
- Updating AppAuth to 3.0
- Raising the various deployment targets to Xcode 27 minimums
- Bump the SPM manifest to 5.7

I've also added `.swiftpm/` to `.gitignore`; whilst not strictly necessary, it is nice in situations like these when a PR can be reviewed, but not yet submitted until a release is fully out.


w-goog:
(Note that this will not be green until https://github.com/google/GTMAppAuth/pull/286)

brnnmrls (approved):

PR #629, opened 2026-08-27
Add non-blocking Xcode 27 preview jobs to CI

w-goog:
Third time is the charm - see https://github.com/google/GTMAppAuth/pull/281


brnnmrls (approved):

.github/workflows/builds.yml
```diff
@@ -18,7 +21,7 @@ jobs:
         os: [macos-15]
   
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
```
brnnmrls: Is there a reason we're not also using v7.0.1 for this one? Also for the ones below.

w-goog: There is not - updated!

PR #632, opened 2026-08-28
Fix SignInSample import + add to CI

w-goog:
This header only existed because use_frameworks smooshes all of the pod headers. SPM doesn't do that, so, the import didn't work when building the SPM sample. I've also added a CI job to build both of them, in order to ensure we don't accidentally break it again.


w-goog:
I believe the action scanner complaints will be addressed by https://github.com/google/GoogleSignIn-iOS/pull/629; if not, I'll address them here as well.

AkshatG6 (approved):

AkshatG6:
I think the changes in #629 should resolve the scanner issues but let's verify that before merging this PR.

PR #636, opened 2026-09-09
Update push and pr notification to use Cards v2.

brnnmrls:
This pull request updates .github/workflows/pr_notification.yml and .github/workflows/push_notification.yml to resolve an issue where notifications rendered as blank message bubbles in Google Chat. These files required updating the the following:
- deprecated Cards v1 schema
- unexpanded environment variables inside single-quoted curl commands
- JSON syntax failures from unescaped quotes or multiline commit messages. 
 
To fix this, both workflows have been migrated to modern Google Chat Cards v2 (cardsV2) utilizing decoratedText and buttonList widgets.


.github/workflows/pr_notification.yml
```diff
@@ -4,97 +4,122 @@ on:
...
                 }
               ]
-            }
-          ]
-        }'
+            }')
+
+          curl --location --request POST "$WEBHOOK_URL" \
+            --header 'Content-Type: application/json; charset=UTF-8' \
+            --data "$PAYLOAD"
```
w-goog, line 124: nit: Why the change to --data instead of --data-raw?

.github/workflows/pr_notification.yml
```diff
@@ -4,97 +4,122 @@ on:
...
+                    ]
+                  }
                 }
               ]
-            }
-          ]
-        }'
+            }')
+
+          curl --location --request POST "$WEBHOOK_URL" \
```
w-goog, line 122: I don't think we want this to fail a PR, but I do think we want it to indicate that something failed. For both curl commands, you could do all of: a) add `--fail-with-body` b) Capture the error output, and echo it in the log and c) add `continue-on-error: true`

.github/workflows/pr_notification.yml
```diff
@@ -4,97 +4,122 @@ on:
...
+                          {
+                            decoratedText: {
+                              topLabel: "Reviewers",
+                              text: (if $reviewers != "" then "- " + $reviewers else "None" end)
+                            }
+                          },
+                          {
+                            decoratedText: {
+                              topLabel: "Labels",
+                              text: (if $labels != "" then "- " + $labels else "None" end)
```
w-goog, line 97: Please do a little helper function to add this guard on every `text` in both files - for example, I think it's _unlikely_ that "github.event.head_commit.author.username" will be nil, but it is possible, which would cause this to fail. `text` is required by v2.

.github/workflows/pr_notification.yml
```diff
@@ -4,97 +4,122 @@ on:
...
-            {
-              "header": {
-                "title": "Pull request notification",
-                "subtitle": "Pull request: #${{ github.event.pull_request.number }}"
-              },
-              "sections": [
+      - name: Google Chat Notification
+        shell: bash
+        env:
+          WEBHOOK_URL: ${{ secrets.WEBHOOK_URL }}
```
w-goog, line 25: Since this will always be nil on PRs that come in from forks, can you add a quick check for `WEBHOOK_URL` that will skip the workflow if nil?

w-goog (approved):
Approved assuming feedback is addressed :)

.github/workflows/pr_notification.yml
brnnmrls: ah, good catch, no reason, reverted this back to --data-raw.

.github/workflows/pr_notification.yml
brnnmrls: Added to both files!

.github/workflows/pr_notification.yml
brnnmrls: Done in both files.

.github/workflows/pr_notification.yml
brnnmrls: Done in both files.

